PR #119, opened 2021-06-30
IOS-6842 Update font weights

jmpg93:
I'm updating font weights as requested by this ticket [IOS-6842](https://jira.tid.es/browse/IOS-6842) for the following components:
- Rowlist
- Cards
- Highligted card
- Emtpy state card


Mistica/Source/Components/Feedback/FeedbackView.swift
```diff
@@ -65,7 +65,7 @@ public class FeedbackView: UIView {
     private lazy var subtitleLabel: UILabel = {
         let label = UILabel()
         label.text = subtitle
-        label.font = .textPreset4(weight: .light)
+        label.font = .textPreset4(weight: .regular)
```
DavidMarinCalleja, line 68: The ticket only says title and does not talk about FeedbackView. has the requirement changed?

DavidMarinCalleja (dismissed):

.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ jobs:
       - name: Clean tmp directory
         run: rm -rf tmp/*
       
+      - name: Install rbenv
+        run: brew install rbenv
+
```
jmpg93, line 19: Needed as a part of the ci flow. Note that brew does not install the package if it's already installed.

Mistica/Source/Components/Feedback/FeedbackView.swift
jmpg93: You are right. I thought it was part of the emtpy state, but I think that is a component we haven't made yet. 
Can you confirm @yceballost?

DavidMarinCalleja (approved):

.github/workflows/build-mistica-catalog.yml
```diff
@@ -35,13 +38,16 @@ jobs:
...
+        run: echo "~/.rbenv/shims" >> $GITHUB_PATH
+
       - name: Setup Ruby
         run: |
-          export RBENV_ROOT=/usr/local/rbenv
-          export PATH="${RBENV_ROOT}/bin:$PATH"
           rbenv install 2.6.0 -s
           rbenv local 2.6.0
+          which ruby
+          rbenv which ruby
```
jmpg93, line 50: I'm using the same set up as in CI.yaml

fastlane/Fastfile
```diff
@@ -64,6 +64,12 @@ platform :ios do
         export_team_id: "JA666WMX28",
         output_directory: BUILD_OUTPUT_PATH,
         export_method: 'enterprise',
+        export_options: {
+          signingStyle: "manual",
+          provisioningProfiles: {
+            "com.telefonica.mistica.enterprise": "mistica-app" 
+          }
+        },
```
jmpg93, line 72: gym is not able to detect which profile should be used, even tho there is just one profile. As a part of the ci flow, the profile is uninstalled and installed again every time. I've noticed that the keychain keeps duplicating the same profile, when you try to remove it, the system is not able to find the profile. 

It may be a bug when uninstalling the profile from GitHub actions in Big Sur. I will keep investigating, but for now I'm adding the export option explicitly.

jmpg93:
Sorry guys! I've updated the PR several times, but the build generation was broken since the slave 5 was used as a runner of github actions. I've made a few changes to fix the actions. I've also update the catalog (cc: @yceballost): 
mistica-catalog-download-qr

Mistica/Source/Components/Feedback/FeedbackView.swift
yceballost: Empty State card is a different component! haha don't change feedback screen titles to regular

DavidMarinCalleja (approved):

pbartolome (approved):

yceballost (approved):

PR #131, opened 2021-08-31
fix(SegmentSelector): Fix segment selection after theme changes

jmpg93:
I'm just adding a few tests to the SegmentSelector. I've noticed we had a bug when the userInterfaceStyle changes, so I've fixed it.

I've also made a tiny change to the checkbox. There is no ticket for this. It's a fix for something I noticed while using the app.


pbartolome (approved):

jmbrocal (approved):
**LGTM**

PR #160, opened 2022-01-26
fix(EmptyState): Missing accesiblity IDS in Empty state. Fix wrong color on InputField borders - SMARTWIFI-3416

pbartolome:
Appium, wasn't able to see elements in the EmptyState as the accesibility elements were being overriden with the `accessibilityElements` array. With these fixes now Appium can detect all elements inside the EmptyState, and voiceover still reads the text as a single element (it doesn't jump from one label to the next)

image

- Bonus fix: `InputField.swift` was using the incorrect color for the border

This code has already been used to generate a "maintenance" mistica version https://github.com/Telefonica/mistica-ios/releases/tag/v14.2.1, this PR is to add the fix also in `main`

### DIFF
--- a/Mistica/Source/Components/EmptyState/EmptyState.swift
+++ b/Mistica/Source/Components/EmptyState/EmptyState.swift
@@ -28,18 +28,6 @@ public class EmptyState: UIView {
         }
     }
 
-    override public var accessibilityElements: [Any]? {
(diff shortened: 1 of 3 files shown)


pbartolome:
this PR is incorrect, I will create a new one

PR #161, opened 2022-01-26
fix(EmptyState): Missing accesiblity IDS in Empty state. Fix wrong color on InputField borders - SMARTWIFI-3416

pbartolome:
Appium, wasn't able to see elements in the EmptyState as the accesibility elements were being overriden with the `accessibilityElements` array. With these fixes now Appium can detect all elements inside the EmptyState, and voiceover still reads the text as a single element (it doesn't jump from one label to the next)

image

- Bonus fix: `InputField.swift` was using the incorrect color for the border

This code will be used to generate a "maintenance" mistica version v14.2.1, this PR is to add the fix also in `main`


Mistica/Source/Components/EmptyState/Internals/EmptyStateContentBase.swift
```diff
@@ -113,6 +104,7 @@ extension EmptyStateContentBase {
             iconImage.accessibilityIdentifier
         }
         set {
+            iconImage.isAccessibilityElement = true
```
DavidMarinCalleja, line 107: I didn't know this property, good work!

DavidMarinCalleja (approved):

alexanegon (approved):

PR #163, opened 2022-02-09
feat(MediaCard): added customizable color to pretitle

Gomay88:
https://jira.tid.es/browse/IOS-6632

Added a var so we will be able to modify de pretitle color, also added 4 different colours to try in Mistica Catalog

Simulator Screen Recording - iPhone 13 Pro - 2022-02-09 at 09 41 59


DavidMarinCalleja (approved):
review the spec in Zeplin https://app.zeplin.io/project/5d653c69f828bf7299c551c1/screen/5fd9ef3349a5fc7fd45b8fa9 Maybe this component is already migrated to Figma.

Mistica/Source/Components/Cards/MediaCard.swift
```diff
@@ -16,6 +16,7 @@ public struct MediaCardConfiguration {
     let headline: String?
     let title: String?
     let pretitle: String?
+    let pretitleColor: UIColor?
```
DavidMarinCalleja, line 19: mmmh I'm not sure if we can pass a color directly to the component but rather indicate a semantic color. Because I imagine that each Brand will have its own color for that component and that it will also switch between light and dark mode.

DavidMarinCalleja (changes requested):

Gomay88:
I'm talking with Hector about this ticket and maybe it won't have to apply.

Mistica/Source/Components/Cards/MediaCard.swift
jmpg93: Something like enum PretitleStyle { case normal; case highlighted }

PR #174, opened 2022-04-01
fix(Feedback): Accesibility identifiers for feedback

Gomay88:
Add identifiers to feedback view as requested:

(default IDs of Mistica Feedback screen component)
icon
title
subtitle
first_button

### DIFF
--- a/Mistica/Source/Components/Feedback/FeedbackView.swift
+++ b/Mistica/Source/Components/Feedback/FeedbackView.swift
@@ -51,9 +51,20 @@ public class FeedbackView: UIView {
         animation.contentMode = .scaleAspectFit
         animation.loopMode = .playOnce
         animation.isUserInteractionEnabled = false
-        animation.isAccessibilityElement = true
         return animation
     }()
+
+    var iconAccessibilityIdentifier: String? {
+        get {
+            icon.accessibilityIdentifier ?? animatedIcon.accessibilityIdentifier
+        }
+        set {
+            icon.isAccessibilityElement = true
+            icon.accessibilityIdentifier = newValue
+            animatedIcon.isAccessibilityElement = true
+            animatedIcon.accessibilityIdentifier = newValue
+        }
+    }
 
     private lazy var titleLabel: UILabel = {
         let label = UILabel()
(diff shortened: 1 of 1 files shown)


Mistica/Source/Components/Feedback/FeedbackView.swift
```diff
@@ -54,24 +54,58 @@ public class FeedbackView: UIView {
         return animation
     }()
 
+    var iconAccessibilityIdentifier: String? {
```
cbarcom, line 57: in other components of Mistica, it has been used assetAccessibilityIdentifier naming. What do you think of using the same name?

Mistica/Source/Components/Feedback/FeedbackView.swift
```diff
@@ -54,24 +54,58 @@ public class FeedbackView: UIView {
...
+        label.isAccessibilityElement = true
         return label
     }()
 
+    var titleAccessibilityIdentifier: String? {
+        get {
+            titleLabel.accessibilityIdentifier
+        }
+        set {
+            titleLabel.isAccessibilityElement = true
```
cbarcom, line 84: is this needed? it has been set already during its creation

Mistica/Source/Components/Feedback/FeedbackView.swift
```diff
@@ -54,24 +54,58 @@ public class FeedbackView: UIView {
...
+        label.isAccessibilityElement = true
         return label
     }()
 
+    var subtitleAccessibilityIdentifier: String? {
+        get {
+            subtitleLabel.accessibilityIdentifier
+        }
+        set {
+            subtitleLabel.isAccessibilityElement = true
```
cbarcom, line 104: same here

Mistica/Source/Components/Feedback/FeedbackView.swift
DavidMarinCalleja: +1

Mistica/Source/Components/Feedback/FeedbackView.swift
```diff
@@ -54,24 +54,58 @@ public class FeedbackView: UIView {
         return animation
     }()
 
+    var iconAccessibilityIdentifier: String? {
+        get {
+            icon.accessibilityIdentifier ?? animatedIcon.accessibilityIdentifier
```
DavidMarinCalleja, line 59: since this sets a value in `accessibilityIdentifier` for _icon_ and _animatedIcon_ at the same time, it may only be necessary to return one value because the second one will have the same value.
```suggestion
            icon.accessibilityIdentifier
```

DavidMarinCalleja (changes requested):

cbarcom (approved):

DavidMarinCalleja (approved):

PR #195, opened 2022-07-20
fix(TitleView): Make TitleView and TitleViewTest public adding them i…

idenjoe:
Fix previous commit wrong message to be able to generate v17.0.1


DavidMarinCalleja (dismissed):

amegias (dismissed):
thx!

PR #199, opened 2022-08-01
fix(Prominent): Fix prominent theme colors for Movistar

jmpg93:
Hi there!

I'm fixing prominent variant colors for Movistar using the [latests definition of design](https://github.com/Telefonica/mistica-design/blob/5cfbe888c44a25145c136398a56a2726d254ba1e/tokens/movistar.json#L846).
It broke while aligning the colors of mistica with android and web a few weeks ago.

### DIFF
--- a/Mistica/Source/Colors/MovistarColorPalette.swift
+++ b/Mistica/Source/Colors/MovistarColorPalette.swift
@@ -17,10 +17,10 @@ struct MovistarColors: Colors {
     let backgroundOverlay = MovistarColorPalette.grey6.withAlphaComponent(0.6) | MovistarColorPalette.darkModeGrey.withAlphaComponent(0.8)
     let backgroundSkeleton = MovistarColorPalette.grey1 | MovistarColorPalette.grey6
     let backgroundSkeletonInverse = MovistarColorPalette.movistarBlue55 | MovistarColorPalette.grey6
-    let navigationBarBackground = MovistarColorPalette.movistarBlue | MovistarColorPalette.darkModeBlack
+    let navigationBarBackground = MovistarColorPalette.movistarBlueWithProminentVariant | MovistarColorPalette.darkModeBlack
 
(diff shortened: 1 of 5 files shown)


Mistica/Source/Colors/MovistarColorPalette.swift
```diff
@@ -17,10 +17,10 @@ struct MovistarColors: Colors {
     let backgroundOverlay = MovistarColorPalette.grey6.withAlphaComponent(0.6) | MovistarColorPalette.darkModeGrey.withAlphaComponent(0.8)
     let backgroundSkeleton = MovistarColorPalette.grey1 | MovistarColorPalette.grey6
     let backgroundSkeletonInverse = MovistarColorPalette.movistarBlue55 | MovistarColorPalette.grey6
-    let navigationBarBackground = MovistarColorPalette.movistarBlue | MovistarColorPalette.darkModeBlack
+    var navigationBarBackground: UIColor { MovistarColorPalette.movistarBlueWithProminentVariant | MovistarColorPalette.darkModeBlack }
```
jmpg93, line 20: We need to use computed properties. Otherwise, the prominent color will be wrong.

salavert: should we state some kind of premises so other will have this problem into account in the future?

jmpg93: What do you mean?

alexanegon (approved):

Mistica/Source/Colors/MovistarColorPalette.swift
jmpg93: Talked offline. The best option for this to not happen again is to create tests for prominent. I've create this ticket to https://jira.tid.es/browse/IOS-7587 tackle it down at some point.

salavert (approved):

PR #209, opened 2022-08-12
fix(Version): Fix mistica version and catalog version

jmpg93:
Hi there!

I'm fixing the mistica catalog version and mistica version.
I'm also updating the script to update version which seems to be working wrong.


DavidMarinCalleja (approved):

jmpg93:
Sorry @DavidMarinCalleja, I've stale your review. I've added dark mode assets as requested by @yceballost

MisticaCatalog/Source/ColorsView.swift
```diff
@@ -95,7 +95,7 @@ struct Searchable: ViewModifier {
 
 extension UIImage {
     func bordered(borderWidth: CGFloat = 1, color: UIColor) -> UIImage {
-        UIGraphicsBeginImageContext(size)
+        UIGraphicsBeginImageContextWithOptions(size, false, UIScreen.main.scale)
```
DavidMarinCalleja, line 98: Just out of curiosity, why do we need to set this option (`UIScreen.main.scale`) now?

DavidMarinCalleja (approved):

MisticaCatalog/Source/ColorsView.swift
jmpg93: Because we need to take the scale into account for core graphics to render the image properly. I didn't notice until installing in a real device.

jmbrocal (approved):

PR #210, opened 2022-08-24
Link checker action

yceballost:
You can trigger this action manually to know if there exist links broken in this repo. 

At this moment there is a lot through the markdown documentation
https://github.com/Telefonica/mistica-ios/actions/runs/2919640915

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -1,0 +1,24 @@
+name: Links checker
+
+on:
+  repository_dispatch:
+  workflow_dispatch:
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,23 @@
+name: Links checker
+
+on:
+  workflow_dispatch:
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
```
amegias, line 8: should we use `self-host` @Yermanaco ?

DavidMarinCalleja:
So far we haven't experienced this broken link problem and we don't see the need to add this feature. If the need arises in the future we will talk about it. Thanks for your interest, @yceballost .

PR #211, opened 2022-08-24
fix(Assets): IOS-7644 Exclude common assets from  protocol

amegias:
https://jira.tid.es/browse/IOS-7644

I want to use this dependency (instead of mistica-ios-swiftui) in Zeus project. I've noticed that `MisticaBrandAssets` protocol includes some **common assets** plus the _branded_ ones. This protocol allows to customize the _branded_ assets when we use a custom brand instead of the existent ones (movistar, etc) so having to implement common resources every time we create a custom brand does not make sense.

What did I do??? I've just removed the common assets since they will be shared internally for all our brands included the custom ones.

I've not considered this a breaking change since the old clients will just to update and remove (if they want; the should) the excluded properties in the protocol.

### DIFF
--- a/Sources/MisticaCommon/Assets/AssetToolkit+UIImage.swift
+++ b/Sources/MisticaCommon/Assets/AssetToolkit+UIImage.swift
@@ -11,31 +11,31 @@ import UIKit
 
 public extension UIImage {
     static var arrowRight: UIImage {
-        MisticaConfig.currentBrandAssets.arrowRight
+        UIImage(named: "icn_arrow_right", type: .common)!
     }
 
     static var checkmarkIcon: UIImage {
-        MisticaConfig.currentBrandAssets.checkmarkIcon
+        UIImage(named: "icn_checkbox_check", type: .common)!
     }
 
     static var eyeEnabled: UIImage {
-        MisticaConfig.currentBrandAssets.eyeEnabled
+        UIImage(named: "icn_eye_enabled", type: .common)!
     }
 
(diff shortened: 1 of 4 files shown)


Sources/MisticaCommon/Assets/AssetToolkit+UIImage.swift
```diff
@@ -11,38 +11,38 @@ import UIKit
...
+        UIImage(named: "icn_calendar", type: .common)!
     }
 
     static var iconNotificationInfo: UIImage? {
         MisticaConfig.currentBrandAssets.iconNotificationInfo
     }
 
     static var closeButtonBlackSmallIcon: UIImage {
-        MisticaConfig.currentBrandAssets.closeButtonBlackSmallIcon
+        UIImage(named: "icn_close_black_small", type: .common)!
```
amegias, line 46: These assets shouldn't belong to the protocol since the protocol is used to inject the **_branded_ assets** and they are common assets

Sources/MisticaCommon/Assets/DefaultMisticaBrandAssets.swift
```diff
@@ -14,42 +14,10 @@ public struct DefaultMisticaBrandAssets: MisticaBrandAssets {
         // Do nothing
     }
```
amegias, line 16: only **.branded**

alexanegon (approved):

jmpg93 (approved):

PR #216, opened 2022-09-16
Use Xcode 13.4.1 to improve job stability

jmpg93:
There is a bug in [Xcode 13.2](https://developer.apple.com/forums/thread/696504) that is not able to resolve SPM dependencies with the current project configuration (after merging mistica-swiftui inside this repo) and makes de CI job to fail.

As our machines already support Xcode 13.4.1 we can use it for now.


.github/workflows/build-mistica-catalog.yml
```diff
@@ -45,6 +45,4 @@ jobs:
 
       - name: Upload ipa to AppCenter
         run: npx -p appcenter-cli appcenter distribute release -a Tuenti-Organization/Mistica-iOS -f ./build/ios.ipa -g Public --token ${{ secrets.APPCENTER_API_TOKEN }}
-        
-      - name: Upload dsyms to AppCenter
-        run: npx -p appcenter-cli appcenter crashes upload-symbols -a Tuenti-Organization/Mistica-iOS -s ./build/ios.dSYM --token ${{ secrets.APPCENTER_API_TOKEN }}
```
jmpg93, line 50: We can skip this to make builds faster.

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: help setup format test simulator archive export clean
 
 # Simulator
-OS_VERSION := 15.2
+OS_VERSION := 15.5
```
jmpg93, line 4: The makefile is able to autogenerate the simulator if it doesn't exist.

amegias (approved):
Do not forget to execute `build-mistica-catalog.yml` before integrating this ;)

alejandroruizponce (approved):
Great!

.github/workflows/build-mistica-catalog.yml
jmpg93: [](url)

> Do not forget to execute `build-mistica-catalog.yml` before integrating this ;)

Deploy working fine https://github.com/Telefonica/mistica-ios/actions/runs/3080655843/jobs/4978259619

PR #217, opened 2022-09-20
IOS-7705 Allow icons from url in Sheet

alejandroruizponce:
Hi there!

This PR is a hotfix to map the icon parameter of the rows in the Sheet component. Until now it was thought that it was not going to be necessary to include it but we already have the definition and it is required for 13.8.

This parameter will come with two icon urls, one for the light mode and one for the dark mode.

### DIFF
--- a/Sources/Mistica/Components/Sheet/View/SheetView.swift
+++ b/Sources/Mistica/Components/Sheet/View/SheetView.swift
@@ -135,12 +135,17 @@ private extension SheetView {
                     imageView.heightAnchor.constraint(equalToConstant: 24.0).isActive = true
                     imageView.widthAnchor.constraint(equalToConstant: 24.0).isActive = true
                     imageView.contentMode = .scaleAspectFit
-                    if let urlToDisplay = icon.urlToDisplay,
-                       let url = URL(string: urlToDisplay),
-                       let data = try? Data(contentsOf: url) {
-                        imageView.image = UIImage(data: data)
-                    }
-
+					if let url = URL(string: icon.url),
+						let data = try? Data(contentsOf: url) {
+						let lightImage = UIImage(data: data)
+						if let iconDark = icon.urlDark,
+						   let urlDark = URL(string: iconDark),
+						   let dataDark = try? Data(contentsOf: urlDark),
+						   let darkImage = UIImage(data: dataDark) {
+								lightImage?.imageAsset?.register(darkImage, with: .init(userInterfaceStyle: .dark))
+						}
+						imageView.image = lightImage
+					}
                     itemStackView.addArrangedSubview(imageView)
                 }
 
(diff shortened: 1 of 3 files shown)


Sources/Mistica/Components/Sheet/View/SheetView.swift
```diff
@@ -135,7 +135,12 @@ private extension SheetView {
                     imageView.heightAnchor.constraint(equalToConstant: 24.0).isActive = true
                     imageView.widthAnchor.constraint(equalToConstant: 24.0).isActive = true
                     imageView.contentMode = .scaleAspectFit
-                    imageView.image = icon
+                    if let urlToDisplay = icon.urlToDisplay,
+                       let url = URL(string: urlToDisplay),
+                       let data = try? Data(contentsOf: url) {
+                        imageView.image = UIImage(data: data)
```
amegias, line 141: is the `imageView` updated every time the theme (light/dark) changes??? IMO it should

alexanegon, line 141: Maybe we could create a UIImageView subclass that has a `urlToDisplay` property and it is responsible to manage all the download stuff. WDYT?

alejandroruizponce: Fixed!

alejandroruizponce: I have discarded the urlToDisplay, it was necessary to create an image asset to dynamically handle the light interface changes.

amegias (approved):

Sources/Mistica/Components/Sheet/View/SheetView.swift
```diff
@@ -135,7 +135,17 @@ private extension SheetView {
...
+                       let data = try? Data(contentsOf: url) {
+                        let lightImage = UIImage(data: data)
+                        if let iconDark = icon.urlDark,
+                           let urlDark = URL(string: iconDark),
+                           let dataDark = try? Data(contentsOf: urlDark),
+                           let darkImage = UIImage(data: dataDark) {
+                            lightImage?.imageAsset?.register(darkImage, with: .init(userInterfaceStyle: .dark))
+                        }
+                        imageView.image = lightImage
+                    }
```
alexanegon, line 148: I am still thinking that would be better to create a subclass where manage all the `if lets`

alexanegon (approved):

Sources/Mistica/Components/Sheet/View/SheetView.swift
alejandroruizponce: Agree but in the next iteration with the inclusion of tests and accessibility, this component can be refactored in subclasses, but only subclassing a part is not uniform.

PR #221, opened 2022-09-27
IOS-7725 Load icons asynchronously

amegias:
https://jira.tid.es/browse/IOS-7725

The icons are being loaded in a sync flow. It blocks the view creation and then the presentation (See the ticket->video).

I've moved to an async flow to load the icons and then, fix this ticket.

We will work on placeholders, error cases, etc in the future

https://user-images.githubusercontent.com/9945756/192541308-d084e469-4e76-4a26-911c-47769e00c97b.mov

PR #225, opened 2022-10-04
IOS-7731 Fix Sheet component with dismissal bug

alejandroruizponce:
Hi there!

This PR has several objectives that are linked to each other:
- Fix the dismissal bug when we drag down a bit with the Sheet. Taking as reference the Novum code I have been able to refactor my transitions delegate so that it works properly in all cases. It will be necessary to add that delegate to the Novum code as well so that everything works correctly.
- This refactor has entailed some modifications in the Sheet view as well as a cleaning and reordering of the code.
- Having modified the view and the way of instantiating it, I have preferred to unify in this PR the UI tests, since those of the previous PR would have been obsolete for the new view.

I put @yceballost in review also in case you want to review again the behavior of the component, there have been several changes, among them, the fixing of the bug.


Sources/Mistica/Components/Sheet/View/SheetViewController.swift
```diff
@@ -6,42 +6,390 @@
...
+            contentSelected.append(.init(id: sender.contentId, selected: [sender.itemId]))
+            content.filter { $0.key != sender.itemId }.forEach { $1.isActivated = false }
+            item.isActivated = true
+            if !isDismissing && sender.autoSubmit {
+                isDismissing = true
+                sheetSelectionResponse.action = .submit
+                sheetSelectionResponse.selectedIds = contentSelected
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+                    self.dismiss(animated: true)
+                }
```
amegias, line 391: why is it needed??? This delay is ugly...

https://user-images.githubusercontent.com/9945756/193849381-39fb30e3-3dcf-4758-88bf-5f8b175df12e.mov

Sources/Mistica/Components/Sheet/View/SheetViewController.swift
```diff
@@ -6,42 +6,390 @@
...
 
-    override public func viewWillDisappear(_ animated: Bool) {
-        super.viewWillDisappear(animated)
-        if isBeingDismissed {
-            completionHandler?(sheetView.sheetSelectionResponse)
+    @objc private func selectItem(_ sender: ItemInformationTapGesture) {
+        if let content = contentInfo[sender.contentId], let item = content[sender.itemId] {
+            contentSelected.append(.init(id: sender.contentId, selected: [sender.itemId]))
+            content.filter { $0.key != sender.itemId }.forEach { $1.isActivated = false }
+            item.isActivated = true
```
amegias, line 384:
```suggestion
            content.forEach { $1.isActivated = $1 == item }
```

amegias (dismissed):
Really good job, Álex!!!!!
Thanks for taking care of the component :)

Sources/Mistica/Components/Sheet/View/SheetViewController.swift
alejandroruizponce: I think that at the UX level a little delay is nice, so that the user can see which option has been selected before dismiss it. I have reduced it to 0.25, I think it looks better now.

Sources/Mistica/Components/Sheet/View/SheetViewController.swift
alejandroruizponce: You're very right! Thanks!

amegias (approved):

amegias (approved):

Sources/Mistica/Components/Sheet/View/SheetViewController.swift
jmbrocal: I agree with @amegias this delay is ugly for me too. This kind of decision must be defined by UX/UI team.

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogSheetViewController.swift
```diff
@@ -187,11 +187,12 @@ extension UICatalogSheetViewController: UITableViewDataSource, UITableViewDelega
             content: [content]
         )
 
-        let viewController = SheetViewController(configuration: configuration) { sheetResponse in
+        let viewController = SheetViewController(configuration: configuration, completionHandler: { sheetResponse in
```
jmbrocal, line 190: why do we need this change? is it due swift format?

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogSheetViewController.swift
```diff
@@ -201,15 +202,18 @@ extension UICatalogSheetViewController: UITableViewDataSource, UITableViewDelega
...
     var sheetTitle: String? {
-        titleCell.textField.text
+        guard let title = titleCell.textField.text, !title.isEmpty else { return nil }
+        return title
     }
 
     var sheetSubtitle: String? {
-        subtitleCell.textField.text
+        guard let subtitle = subtitleCell.textField.text, !subtitle.isEmpty else { return nil }
+        return subtitle
```
jmbrocal, line 211: The same than above

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogSheetViewController.swift
```diff
@@ -201,15 +202,18 @@ extension UICatalogSheetViewController: UITableViewDataSource, UITableViewDelega
...
     }
 
     var sheetDescription: String? {
-        descriptionCell.textField.text
+        guard let description = descriptionCell.textField.text, !description.isEmpty else { return nil }
+        return description
```
jmbrocal, line 216: The same than above

Sources/Mistica/Components/Sheet/SheetTransition/SheetPresentationController.swift
```diff
@@ -47,242 +84,47 @@ final class SheetPresentationController: UIPresentationController {
...
-        }
-
-        overlayView.frame = containerView.bounds
-
-        guard let presentedView = presentedView else {
-            return
-        }
+    var maxTopOffset: CGFloat {
+        let safeTopInset: CGFloat
+        safeTopInset = containerView?.safeAreaInsets.top ?? 0
```
jmbrocal, line 92:
```suggestion
        let safeTopInset: CGFloat = containerView?.safeAreaInsets.top ?? 0
```

I am not sure if we need the type here.

Sources/Mistica/Components/Sheet/SheetTransition/SheetPresentationController.swift
```diff
@@ -47,242 +84,47 @@ final class SheetPresentationController: UIPresentationController {
...
-        if scrollView.contentOffset.y < .zero {
-            let originalOffset = CGPoint(x: scrollView.contentOffset.x, y: -scrollView.safeAreaInsets.top)
-            scrollView.setContentOffset(originalOffset, animated: false)
+    func animateAlongsideTransitionIfPossible(_ animation: @escaping () -> Void) {
+        if let coordinator = presentedViewController.transitionCoordinator {
+            coordinator.animate(alongsideTransition: { _ in
+                animation()
+            })
+        } else {
+            animation()
```
jmbrocal, line 123: Why do we need this function? The presentedViewController is the sheet and the we always set a transition coordinator, maybe here we should add a guard with an log message or fatal error?

Sources/Mistica/Components/Sheet/SheetTransition/SheetTransitioningDelegate.swift
```diff
@@ -8,43 +8,74 @@
 
 import UIKit
 
+public class SheetTransitioningDelegate: NSObject {
```
jmbrocal, line 11: Need to be public?

Sources/Mistica/Components/Sheet/SheetTransition/SheetTransitioningDelegate.swift
```diff
@@ -8,43 +8,74 @@
...
+
+    public func sheetViewControllerDidEndDragging(_ selectorViewController: SheetViewController) {
+        finishDismiss()
+    }
+}
+
+private extension SheetTransitioningDelegate {
+    func startDismiss(_ viewController: UIViewController) {
+        isInteractionInProgress = true
+        viewController.dismiss(animated: true, completion: nil)
```
jmbrocal, line 65: Should we set to false `isInteractionInProgress` when the completion block is called?

Sources/Mistica/Components/Sheet/SheetTransition/SheetPresentationController.swift
```diff
@@ -47,242 +84,47 @@ final class SheetPresentationController: UIPresentationController {
...
-    /// - Parameter translation: The translation of the presented view used to calculate the progress.
-    private func updateTransitionProgress(for translation: CGPoint) {
-        guard let transitioningDelegate = transitioningDelegate else {
-            return
-        }
+    func setUpDimmingView() {
+        dimmingView.translatesAutoresizingMaskIntoConstraints = false
+        dimmingView.backgroundColor = .backgroundOverlay
+        dimmingView.alpha = 0.0
+        dimmingView.accessibilityIdentifier = "dimmingView"
```
jmbrocal, line 105: Are we using this identifier?

Tests/MisticaTests/UI/SheetTests.swift
```diff
@@ -0,0 +1,292 @@
...
+        var rows: [SheetListRow] = []
+        for index in 1 ... 10 {
+            let item = SheetListRow(
+                id: index.description,
+                title: "Element \(index)",
+                description: "Description",
+                icon: nil
+            )
+            rows.append(item)
+        }
```
jmbrocal, line 43: This code is repeated in other tests, maybe you can create an AnyValues enum and adding this there to be reused

jmbrocal, line 43: This code is repeated in other tests, maybe you can create an AnyValues enum and adding this there to be reused

Tests/MisticaTests/UI/SheetTests.swift
```diff
@@ -0,0 +1,292 @@
...
+        let content = SheetList(
+            id: UUID().uuidString,
+            type: "LIST",
+            listType: "SINGLE_SELECTION",
+            autoSubmit: true,
+            selectedId: "1",
+            items: rows
+        )
```
jmbrocal, line 51: The same for this content

Tests/MisticaTests/UI/SheetTests.swift
```diff
@@ -0,0 +1,292 @@
...
+        assertSnapshotForAllBrandsAndStyles(
+            as: .image(on: .iPhoneSe),
+            viewBuilder: sheetView(
+                title: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+                subtitle: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+                description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+                content: []
+            )
+        )
+    }
```
jmbrocal, line 31: This test is not needed because the component does not allow this case.

Tests/MisticaTests/UI/SheetTests.swift
```diff
@@ -0,0 +1,292 @@
...
+//  Made with ❤️ by Novum
+//
+//  Copyright © Telefonica. All rights reserved.
+//
+
+@testable import Mistica
+import SnapshotTesting
+import XCTest
+
+final class SheetTests: XCTestCase {
```
jmbrocal, line 13: IMO this tests are not correct because they are testing the component in isolation but this component is never used alone. It is always shown over a view. The following points of the component are not being tested
- the dimmed view
- the component height based on the content

I think we should use a more real example, with a dummy view controller which is showing the sheet.

WDYT?

jmbrocal:
I found the following issues after testing the component
- It seems like the icons are downloaded in main thread, you can see the following logs
````
[CFNetwork] Synchronous URL loading of https://img.icons8.com/ios/344/bookmark--v1.png should not occur on this application's main thread as it may lead to UI unresponsiveness. Please switch to an asynchronous networking API such as URLSession.
````
Screen Shot 2022-10-05 at 13 56 30

- When the component has no title or subtitle is not keeping the space the space of the title
current state
Screen Shot 2022-10-05 at 13 48 35
Sample from spec
Screen Shot 2022-10-05 at 13 48 35
- The dismiss animation when dismissing the sheet using a pan gesture is not fluid. the component moves down more slowly than my finger.

File: Sources/Mistica/Components/Sheet/View/SheetViewController.swift [hunk 15]
```diff
@@ -6,42 +6,389 @@
...
+                    if let url = URL(string: icon.url),
+                       let data = try? Data(contentsOf: url) {
+                        let lightImage = UIImage(data: data)
+                        if let iconDark = icon.urlDark,
+                           let urlDark = URL(string: iconDark),
+                           let dataDark = try? Data(contentsOf: urlDark),
+                           let darkImage = UIImage(data: dataDark) {
+                            lightImage?.imageAsset?.register(darkImage, with: .init(userInterfaceStyle: .dark))
+                        }
+                        imageView.image = lightImage
```
amegias, line 254: bad merge @alejandroruizponce.
as @jmbrocal said, it's wrong. It's already solved in main, so be careful since it looks an old version of the component....
Solved here: https://github.com/Telefonica/mistica-ios/pull/221

amegias (changes requested):

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogSheetViewController.swift
alejandroruizponce: Nope, back and fixed!

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogSheetViewController.swift
```diff
@@ -201,15 +202,18 @@ extension UICatalogSheetViewController: UITableViewDataSource, UITableViewDelega
 
 private extension UICatalogSheetViewController {
     var sheetTitle: String? {
-        titleCell.textField.text
+        guard let title = titleCell.textField.text, !title.isEmpty else { return nil }
+        return title
```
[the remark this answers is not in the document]
alejandroruizponce: It is a simple way to hide those labels in the view when they are null. What you say would also be valid.

Sources/Mistica/Components/Sheet/SheetTransition/SheetPresentationController.swift
alejandroruizponce: Yes, many lines. Fixed!

Sources/Mistica/Components/Sheet/SheetTransition/SheetPresentationController.swift
alejandroruizponce: I think that returning a return if the coordinator is not present would make sense. It is consistent with the title of the function. Fixed!

Sources/Mistica/Components/Sheet/SheetTransition/SheetTransitioningDelegate.swift
alejandroruizponce: Yes, it has to be public in order to instantiate the delegate from outside.

Sources/Mistica/Components/Sheet/SheetTransition/SheetTransitioningDelegate.swift
alejandroruizponce: It is set to false when dismiss ends.

Sources/Mistica/Components/Sheet/SheetTransition/SheetPresentationController.swift
alejandroruizponce: Nope really, fixed.

Tests/MisticaTests/UI/SheetTests.swift
alejandroruizponce: Yes, you are right, I have created an auxiliary function to reuse code.

Tests/MisticaTests/UI/SheetTests.swift
alejandroruizponce: True, removed.

Sources/Mistica/Components/Sheet/View/SheetViewController.swift
alejandroruizponce: Yes, it was a bad merge. Fixed now!

Tests/MisticaTests/UI/SheetTests.swift
alejandroruizponce: This is something I discussed with @amegias when I did the first tests, but we discarded it because it involved complex changes and the delay of the Sheet animation in appearing gave problems to show it in a basic VC. The snapshot was always done before the Sheet was displayed.

alejandroruizponce:
> I found the following issues after testing the component
> 
> * It seems like the icons are downloaded in main thread, you can see the following logs
> 
> ```
> [CFNetwork] Synchronous URL loading of https://img.icons8.com/ios/344/bookmark--v1.png should not occur on this application's main thread as it may lead to UI unresponsiveness. Please switch to an asynchronous networking API such as URLSession.
> ```
> 
> Screen Shot 2022-10-05 at 13 56 30
> 
> * When the component has no title or subtitle is not keeping the space of the title
>   current state
> 
> Screen Shot 2022-10-05 at 13 48 35
> 
> Sample from spec
> Screen Shot 2022-10-05 at 13 48 35
> 
> - The dismiss animation when dismissing the sheet using a pan gesture is not fluid. the component moves down more slowly than my finger.

When @yceballost  returns from holidays, it would be interesting to get his opinion on the title and the swipe animation. The first seems weird to me that you have to leave that space and the second, about swipe animation, is as it was in Novum animator.

amegias (approved):

Tests/MisticaTests/UI/SheetTests.swift
amegias: AFAIK, we should add a host app to test presentations, etc with VC. So currently we are only able to test views (or static VControllers)

jmbrocal: But for ui testing we do not tests animations because we cant xD. My point is that testing this component using a dummy viewcontroller allows to tests it in deph.

Maybe you have to do some change in code so the component disable animations or makes it animates very fast, but I did it for other components with no issues.

I you were talked about this and decide to go with the proposed approach, I am ok, but IMHO we can do a better job when testing.

yceballost (approved):
like a charm✨

PR #231, opened 2022-11-23
IOS-7902 add a customisable time interval to dismiss

DavidMarinCalleja:
https://jira.tid.es/browse/IOS-7902

This PR consists of adding the possibility to customize the time it takes for the Crouton component to disappear. By default, the time for a Crouton without actions is five seconds, and sometimes that may not seem long enough, for example when showing a warning to the user that he has no connection. 

What I have done is give freedom to the developer to choose that time for any style of Crouton. The design team they have proposed to provide a choice between 5 and 10 seconds.

### DIFF
--- a/MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogCroutonViewController.swift
+++ b/MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogCroutonViewController.swift
@@ -44,11 +44,21 @@ class UICatalogCroutonViewController: UITableViewController {
         cell.textLabel?.text = "Show Crouton"
         return cell
     }()
+    
+    private lazy var croutonDismissIntervalCell: UISegmentedControlTableViewCell = {
+        let cell = UISegmentedControlTableViewCell(reuseIdentifier: "crouton-dismiss-interval")
+        for interval in  CroutonDismissInterval.allCases {
+            cell.segmentedControl.insertSegment(withTitle: "\(interval.rawValue) seconds", at: 0, animated: false)
+        }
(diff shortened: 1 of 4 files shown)


MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogCroutonViewController.swift
```diff
@@ -45,10 +45,20 @@ class UICatalogCroutonViewController: UITableViewController {
...
+        cell.segmentedControl.selectedSegmentIndex = 0
+        return cell
+    }()
+
     private lazy var cells = [
         [titleCell],
         [actionTitleCell],
-        [croutonStyleCell, showCroutonCell]
+        [croutonStyleCell],
+        [croutonDismissIntervalCell, showCroutonCell]
```
DavidMarinCalleja, line 61: I add in the catalogue a UISegmentedControl to be able to change the time.

Sources/Mistica/Components/Crouton/Domain/CroutonConfig.swift
```diff
@@ -14,12 +14,24 @@ import UIKit
...
+    var timeInterval: TimeInterval {
+        TimeInterval(rawValue)
+    }
+}
+
 public struct CroutonConfig {
     let backgroundColor: UIColor
     let textColor: UIColor
     let actionStyle: Button.Style
+    let overrideDismissInterval: TimeInterval?
```
DavidMarinCalleja, line 32: If this value is set, then it overwrites the default value defined by style.

jmpg93 (approved):

cbarcom (approved):

PR #233, opened 2022-12-02
refactor(MisticaButton): Added public access level to MisticaButton and style

salavert:
In order to have custom MisticaButton styles we need to access the initializer

Screenshot 2022-12-02 at 11 41 59

### DIFF
--- a/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
+++ b/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
@@ -13,7 +13,7 @@ public enum ButtonBleedingAlignment {
 }
 
 @available(iOS 13.0, *)
-struct MisticaButton: View {
+public struct MisticaButton: View {
     enum Constants {
         static let minWidth: CGFloat = 76
         static let height: CGFloat = 24
--- a/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
+++ b/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
@@ -58,15 +58,35 @@ public struct MisticaButton: View {
     }
 
     public struct StateStyle {
-        public let textColor: Color
-        public let backgroundColor: Color
-        public let borderColor: Color
+        let textColor: Color
+        let backgroundColor: Color
+        let borderColor: Color
+        
+        public init(
+            textColor: Color,
+            backgroundColor: Color,
+            borderColor: Color
+        ) {
+            self.textColor = textColor
+            self.backgroundColor = backgroundColor
+            self.borderColor = borderColor
+        }
     }
 
     let configuration: ButtonStyle.Configuration
     let style: Style
     let small: Bool
 
+    public init(
+        configuration: ButtonStyle.Configuration,
+        style: Style,
+        small: Bool
+    ) {
+        self.configuration = configuration
+        self.style = style
+        self.small = small
+    }
+    
     @Environment(\.misticaButtonLoadingInfo) private var loadingInfo
     @Environment(\.isEnabled) private var isEnabled
 
--- a/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
+++ b/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
@@ -61,7 +61,7 @@ public struct MisticaButton: View {
         let textColor: Color
         let backgroundColor: Color
         let borderColor: Color
-        
+
         public init(
             textColor: Color,
             backgroundColor: Color,
@@ -86,7 +86,7 @@ public struct MisticaButton: View {
         self.style = style
         self.small = small
     }
-    
+
     @Environment(\.misticaButtonLoadingInfo) private var loadingInfo
     @Environment(\.isEnabled) private var isEnabled
 
--- a/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
+++ b/Sources/MisticaSwiftUI/Components/Button/MisticaButton.swift
@@ -25,11 +25,11 @@ public struct MisticaButton: View {
         static let transitionOffset: CGFloat = 40
     }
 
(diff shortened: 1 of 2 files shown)


pbartolome (approved):
maybe the mistica button should not allow this level of customization... what if this button is something that only Latch uses?, I don't really have a strong opinion on this (as I don't like duplicating the button on Latch), but we usually only allow the customizations defined by the Mistica design team

aweell:
I want to raise my concern about this topic, from the Mística design team we have a strong opinion against open component customization.

* Is the color in this case providing critical information not included in the page text?
* Are actions like "contactar proveedor" critical or destructive so contextual action color is needed?

As defined in the [feedback screen documentation page](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/patterns/feedbacks), the action color is independent of the type of feedback provided (info, error, success). Is the meaning of the action itself which defines the [button variant](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/components/buttons/typology) to use:

For example:

If "Bloquear servicio" is a destructive action that can be easily undone, the correct variant to use would be danger.

**It would be convenient, before merging this PR to have a better understanding of the needs in case it's possible to reach a solution that doesn't imply public access to styles.**

salavert:
> I want to raise my concern about this topic, from the Mística design team we have a strong opinion against open component customization.
> 
> * Is the color in this case providing critical information not included in the page text?
> * Are actions like "contactar proveedor" critical or destructive so contextual action color is needed?
> 
> As defined in the [feedback screen documentation page](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/patterns/feedbacks), the action color is independent of the type of feedback provided (info, error, success). Is the meaning of the action itself which defines the [button variant](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/components/buttons/typology) to use:
> 
> For example:
> 
> If "Bloquear servicio" is a destructive action that can't be easily undone, the correct variant to use would be danger.
> 
> **It would be convenient, before merging this PR to have a better understanding of the needs in case it's possible to reach a solution that doesn't imply public access to styles.**

Why are we having this conversation now and not when Latch Design team proposed this kind of customization or at least when Android implemented them a week ago. I'm tempted to duplicate my own version of MisticaButtons into my project and forget about this PR.

salavert:
> I want to raise my concern about this topic, from the Mística design team we have a strong opinion against open component customization.
> 
> * Is the color in this case providing critical information not included in the page text?
> * Are actions like "contactar proveedor" critical or destructive so contextual action color is needed?
> 
> As defined in the [feedback screen documentation page](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/patterns/feedbacks), the action color is independent of the type of feedback provided (info, error, success). Is the meaning of the action itself which defines the [button variant](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/components/buttons/typology) to use:
> 
> For example:
> 
> If "Bloquear servicio" is a destructive action that can't be easily undone, the correct variant to use would be danger.
> 
> **It would be convenient, before merging this PR to have a better understanding of the needs in case it's possible to reach a solution that doesn't imply public access to styles.**

I will transfer your concerns to the Latch design team so they can evaluate it properly, in the meantime I would like to ask why we should limit the range of colors of the button component, what if error/success/info is not enough? the base component is the tool we need from Mistica into all our projects but we should allow customization to some degree since not all projects are so tied to Mistica design principles.

Btw, in case we don't reach an agreement I´ve already duplicated the Mistica button and created my own button style to meet my team needs but trust me that I'm strongly against this alternative solution.

aweell:
> I will transfer your concerns to the Latch design team so they can evaluate it properly

Please feel free to contact directly if there's any way I can personally help

> why we should limit the range of colors of the button component, what if error/success/info is not enough? the base component is the tool we need from Mistica into all our projects but we should allow customization to some degree since not all projects are so tied to Mistica design principles.

That's, of course, a very valid concern. There should be always red lines in a design system customization, but that doesn't mean the actual lines are placed in the correct position, and each new use case pushes those lines further to meet the different team needs when using Mística. 

The ideal would be: if new contextual colors need to be added in the button component definition, that should be included in the component specs, documented, and implemented for all platforms and brands, so all products can benefit from the improvement. This change can be either done by the Mística team or any other (Following the [contribution model](https://brandfactory.telefonica.com/document/1846#/contribute/how-to-contribute). Since sometimes, certain changes, due to priorities, cannot be made in the expected time). 

Why this is the ideal case? The purpose of Mística, apart from cutting development and design times, is to bring a seamless and consistent user experience across multiple products. An open door for making any change by product in the component definitions can have a negative impact on the latter. If a specific change has been evaluated as an improvement in the user experience and doesn't fit into the Mística component definition (for whatever reason) then it's important to balance how big the improvement is vs the cost of the custom implementation (and maintenance).

If the improvement is worth it there's a reason for the investment. I would not see Mística as something different than a tool that aims to help build digital products rather than guidelines to enforce a specific design for a product.

There's a [discussion forum](https://github.com/Telefonica/mistica-design/discussions) open to anyone that wants to create a new proposal to be evaluated. We are constantly iterating and creating new backlog items based on these proposals. I believe that's the perfect place to challenge the current Mística customization status.

ieduardogf:
Totally agree @aweell. The point here is why Latch needs this kind of extra feedbacks customization that has not been on Mistica before. And if that's the case, it must be evaluated first by Mistica Design team in order to incorporate to the System. 

Any backdoor to Mistica will break the purpose of our Design system really quickly.

salavert:
@aweell our design team has made the decision to follow the Mistica guidelines, as a result, we will need to make some modifications to our feedback component.

We believe that following the Mistica guidelines will help to ensure that our project is consistent and easier to maintain. We will make sure to keep you informed of any updates as we move forward with this project.

Regards.

PR #236, opened 2022-12-21
Add accessibility identifiers for TagView, FeedbackView's link button

cbarcom:
This PR adds :

- Exposes TagView's label accessibility identifier due to there are some cases that might require to change during app execution
- Adds a default identifier for Feedback View's button in case it is a link


Sources/Mistica/Components/Cards/Internals/CardButtonsView.swift
```diff
@@ -45,6 +65,8 @@ extension CardButtons {
         if let primaryAction = primaryButton {
             self.primaryButton.title = primaryAction.title
             self.primaryButton.loadingTitle = primaryAction.loadingTitle
+            self.primaryButton.isAccessibilityElement = true
```
jmbrocal, line 68: I saw we are changing to true a lot of ui components (some of the are even not needed because the default value is true), why do we need these changes? how it is affecting the accessibility reading of the component?

Sources/Mistica/Components/Tag/TagView.swift
```diff
@@ -137,6 +137,15 @@ public class TagView: UIView {
...
     }
 
+    public var labelAccessibilityIdentifier: String? {
+        get {
+            label.accessibilityIdentifier
+        }
+        set {
+            label.accessibilityIdentifier = newValue
+        }
+    }
```
jmbrocal, line 147: Here you should override accesibilityIdentifier

```suggestion
    override public var accessibilityIdentifier: String? {
        get {
            label.accessibilityIdentifier
        }
        set {
            label.accessibilityIdentifier = newValue
        }
    }
```

Sources/Mistica/Components/Cards/MediaCard.swift
```diff
@@ -179,6 +244,17 @@ private extension MediaCard {
...
 
+    func addDefaultAccessibilityIds() {
+        richMediaContainerView.isAccessibilityElement = true
+        richMediaContainerView.accessibilityIdentifier = DefaultIdentifiers.MediaCard.asset
+        baseCardView.contentView.headlineAccessibilityIdentifier = DefaultIdentifiers.MediaCard.headline
+        baseCardView.contentView.titleAccessibilityIdentifier = DefaultIdentifiers.MediaCard.pretitle
+        baseCardView.contentView.subtitleAccessibilityIdentifier = DefaultIdentifiers.MediaCard.title
+        baseCardView.contentView.descriptionAccessibilityIdentifier = DefaultIdentifiers.MediaCard.description
+        baseCardView.buttonsView.primaryButtonAccessibilityIdentifier = DefaultIdentifiers.MediaCard.primaryButton
+        baseCardView.buttonsView.linkButtonAccessibilityIdentifier = DefaultIdentifiers.MediaCard.linkButton
```
jmbrocal, line 255: I think these identifiers can be also configured for other cards types, wdyt?

Sources/Mistica/Components/Tag/TagView.swift
cbarcom: AFAIK, we agreed to give default value to Mistica components but also allow to customize them as it is done in Android.

cbarcom: Not sure if would be the case were tagview needs to have a different Id than its label. In SW they use this ids for automated testing

Sources/Mistica/Components/Cards/MediaCard.swift
cbarcom: I used MediaCard default identifiers as this component items are use to build a MediaCard. I guess there would be other default identifiers for other types of cards but I ignore what the values would be

Sources/Mistica/Components/Cards/Internals/CardButtonsView.swift
cbarcom: There are issues with accessibilityIds and Appium tool. In some cases, like this one, all identifiers of the componentes are merged together and QA can't check if the components of the card are correct. This is an attempt to solve it

jmbrocal: I think this will not solve the issue because as you pointed,  the accessibility inspector read this element is a single unit. I think QA should change how they verify this component or we have to change how the accessibility works for this component, decreasing the usability...

Sources/Mistica/Components/Tag/TagView.swift
jmbrocal: IMO the tag does not need a custom id because we can set a default id based on the style but if we need to set a custom id, I would used my approach overriding the accessibilityIdentifier of the UIView.

Sources/Mistica/Components/Cards/MediaCard.swift
jmbrocal: I think from a point of QA testing view, they should ignore the type of the card. IMHO the test should check the elements but not the type of the elements...

jmpg93 (approved):

Sources/Mistica/Components/Tag/TagView.swift
cbarcom: done

DavidMarinCalleja (approved):

PR #240, opened 2023-01-10
Force semantic release and node versions

cbarcom:
In order not to have problems with semantic release versions and its dependency to Node, both have been forced to use specific versions.

Semantic-release has been forced to use 20.x.x version as before it was using latest version released.
Node has been forced to use 18.0.x version as is the LTS supported by semantic-release 20.0.x 

Captura de Pantalla 2023-01-10 a las 12 20 44

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,12 +21,12 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3.5.0
         with:
-          node-version: 14.17
+          node-version: 18.0
 
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.NOVUM_PRIVATE_REPOS }}
-        run: npx -p @semantic-release/changelog -p @semantic-release/git -p @semantic-release/exec -p semantic-release semantic-release ${{ github.event.inputs.cliArgs }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -21,12 +21,12 @@ jobs:
...
         uses: actions/setup-node@v3.5.0
         with:
-          node-version: 14.17
+          node-version: 18.0
 
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.NOVUM_PRIVATE_REPOS }}
-        run: npx -p @semantic-release/changelog -p @semantic-release/git -p @semantic-release/exec -p semantic-release semantic-release ${{ github.event.inputs.cliArgs }}
+        run: npx -p @semantic-release/changelog -p @semantic-release/git -p @semantic-release/exec -p semantic-release semantic-release@20 ${{ github.event.inputs.cliArgs }}
```
DavidMarinCalleja, line 29: I'm not sure if we need to write the full version `@20.0.0` anyway, the latest version is 20.0.2 and I would add this version
```
semantic-release@20.0.2
```
WDYT?

cbarcom: yes, you're right. If what we want is to fix what version we're using, then we should specify it till the last version component

DavidMarinCalleja (approved):
thanks for fix it

jmpg93 (approved):

PR #200, opened 2022-08-04
feat(SwiftUI): Support Mistica in SwiftUI

jmpg93:
# Overview
Hi there! This PR is about updating mistica to support SwiftUI. The review contains over 1800 new files, but just because I've moved a lot of files and merge another repo here. 

I couldn't make more reviews because I've made 2039 changes along the way, and it was really hard to split in consistent branches.

**I highly suggest to checkout the branch and review it. Otherwise it will be really hard to review. Sorry in advice.**

## Why it is important to have a single repo?
- We only need to maintain a one repo.
- Changes are now more localised and it's easier to move from features from UIKit to SwiftUI (for example, mistica evolution).
- Single source of truth.
- Single catalog.
- Single definition of brands colors and fonts.

## Targets
The repo now has three different targets that can be build separately.
- Mistica: It contains the implementaion of UIKit
- MisticaSwiftUI: It contains the implementaion of SwiftUI
- MisticaCommon: It contains all the common thing including: colors, fonts, assets, some view utils, configuration and brand definition.

When using Mistica, the developer will only need to import `Mistica` or `MisticaSwiftUI`. `MisticaCommon` will be automatically imported, so we don't need to change our corrent WoW.

## Structure Changes
- Replace fastlane by make (I've reused the same make files of mistica-swift-ui)
- Remove Mistica project file (needed to support Carthage). Using SPM carthage is also able to fetch and build the frameworks.
- As there is no mistica project, I've been able to remove carthage dependencies (including the Cartfile).
- Update Catalog app so it can display both UIKit and SwiftUI components.
- Now there are three tests targets. They are all run in the Mistica-Package target test build.
- There is a new MisticaSwiftUI.podspec to support MisticaSwiftUI in Cocoapods.

## Code Changes
- I've made a few changes to the Catalog App to have a single app with two tabs: SwiftUI and UIKit.
- I've made a few changes to use the new colors and fonts of mistica evolution in mistica-swiftui.

## Checks
I've checked that we can work using this new structure in:
- [iphoneapp](https://github.com/Telefonica/iphoneapp/tree/mistica-upgrade) (mistica-ios added as a SPM dependency). 
- [zeus](https://github.com/Telefonica/zeus-ios/tree/mistica-upgrade) (mistica-ios added as a SPM dependency). 
- [aura-ios-container](https://github.com/Telefonica/aura-ios-container/tree/mistica-upgrade) (mistica-ios added as a Carthage dependency). 
[...]

PR #244, opened 2023-02-06
IOS-7854 Remove Cocoa specs

alejandroruizponce:
Hi there!

We had a pending ticket to remove support for using Mistica from Cocoa and Carthage and only support doing it from SPM.

Actually it was almost all removed in this PR from @jmpg93: https://github.com/Telefonica/mistica-ios/pull/200  

But the Cocoa specs were still to be removed. Now everything should be clean so that the Mistica framework is exclusive to SPM.


alexanegon (approved):

jmpg93:
You need to update the scripts that update the version:
https://github.com/Telefonica/mistica-ios/blob/ca5cd24f57ce77eb9868e94a55b1143b666ad788/scripts/update-version.sh#L14

jmpg93 (approved):

PR #246, opened 2023-02-20
Text preset changes

yceballost:
We updated text preset sizes definition
We update the sizes of typography in text-presets for all brands

- text-preset-10 from 32 to 48
- text-preset-5 from 32 to 40
- text-preset-5 from 22 to 20


yceballost:
closing avoiding fork

PR #259, opened 2023-03-13
IOS-8178 Add DataCard font style

alejandroruizponce:
Hi there! 👋

I have added a new fontstyle that we were missing in our list of styles for the DataCard case. Here you can see the style information:
https://www.figma.com/file/aNHyvXe1mqjaelE8XWVQD5/%F0%9F%94%B9-Text-presets-definition?node-id=0-1


Sources/MisticaCommon/Fonts/FontStyle.swift
```diff
@@ -159,7 +165,8 @@ private extension FontStyle {
             return .subheadline
         case .textPreset3:
             return .body
-        case .textPreset4:
+        case .textPreset4,
+             .cardTitlePreset:
```
jmbrocal, line 169: Is this the expected correlation between our text pressets and the iOS dynamic text? I am asking because the figma is not updated
image

Sources/Mistica/Components/Cards/DataCard.swift
```diff
@@ -183,7 +183,7 @@ private extension DataCard {
 
         cardBaseView.contentView.headlineTopSpacing = 8
 
-        cardBaseView.contentView.titleLabel.font = .textPreset4(weight: .medium)
+        cardBaseView.contentView.titleLabel.font = .cardTitlePreset(weight: .bold)
```
DavidMarinCalleja, line 186: Is it necessary to indicate the width? according to Figma it is a brand dependent property, isn't it?

Sources/MisticaCommon/Fonts/FontStyle.swift
alejandroruizponce: Not updated but it's confirmed by @yceballost

jmbrocal (approved):

Sources/MisticaCommon/Styles/BrandStyle.swift
```diff
@@ -55,7 +55,7 @@ public enum BrandStyle {
         [.movistar, .vivo, .o2, .o2Classic, .blau]
     }
 
-    public var usesBoldForTextPreset5AndAbove: Bool {
+    public var isMovistarBrand: Bool {
```
DavidMarinCalleja, line 58: mmh I'm not sure if this name is better. Although the styles depend on each OB, I would not use the name in the name of a function. For example, if the Movistar flag does not exist tomorrow, we will be coupling our method to an implementation detail. 

WDYT guys?

alejandroruizponce: Yes, really this fix must be provisional, the problem of the fonts in each brand must be handled with another better solution in a new PR. The problem that you comment with the name could also be applied to the one that was, if tomorrow instead of being from textPreset5 up occurs only in textPreset9, the name of the function would not be consistent with what we have. Anyway, I have no problem to rename it, I just didn't like the previous name and I wanted to reuse it for cardpreset.

DavidMarinCalleja (approved):

PR #266, opened 2023-03-17
IOS-8208 Add MisticaCatalog config to be pushed with the version bumped

amegias:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-8208

## 🥅 **What's the goal?**
Every release, the Mistica Catalog must be bumped as Mistica already does. Currently, Mistica Catalog version is not pushed.
Captura de pantalla 2023-03-17 a las 12 34 03

## 🚧 **How do we do it?**
The problem is here:
https://github.com/Telefonica/mistica-ios/blob/274db414e3e1dee34e4cb3232b01f57d452cffe7/.releaserc#L9-L11

Every time the semantic-release is executed, it modifies both xcconfigs (Catalog & Mistica), bumping to the next version.
The problem is the Catalog xcconfig is not included there, then it is not pushed :(
See https://github.com/Telefonica/mistica-ios/actions/runs/4446829704/jobs/7807625092#step:4:135 which found only two files (Changelog.md & Mistica.xcconfig).

I've just added the Catalog config.

You can check that `./scripts/update-version.sh 1.2.3` is working properly, modifying both files. We'll check it in the next release if they are commited & pushed.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,23 @@
+## 🎟️ **Jira ticket**
+<!-- Add here the ticket number: [IOS-XXXX](https://jira.tid.es/browse/IOS-XXX) Blah Blah -->
+
+## 🥅 **What's the goal?**
+<!--
+Here is a description of what this pull request is about, with a brief text of the functionality or bug to be fixed. Use the following points to help you:
+- Attach screenshots or videos to describe what you want to fix.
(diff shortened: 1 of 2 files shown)


DavidMarinCalleja (approved):
thanks for fix it

idenjoe (approved):

DavidMarinCalleja (approved):

PR #267, opened 2023-03-31
IOS-8217 Add Brandeable FontWeights

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-8217

## 🥅 **What's the goal?**
The problem is that we are currently not handling correctly the different FontWeights (bold, medium, light...) that each brand has for each TextPreset we use. In a brand like Movistar the weight for TextPreset5 can be bold but for O2 it can be light. We have to handle it appropriately as we already do with colors or assets.

## 🚧 **How do we do it?**
For them we have proposed the following approach that you can see documented in Confluence where we explain the steps followed to handle the different types of FontWeights that currently exist in Design and that we must be able to handle from Mistica:

https://confluence.tid.es/display/CTO/%5BiOS%5D+Mistica+Brandeable+Font+Weights

You can load from Mistica Catalog the brand you want by setting it in AppDelegate and see the differences with the rest of the brands in their FontWeights.


Sources/MisticaCommon/Fonts/MisticaFontWeights.swift
```diff
@@ -0,0 +1,53 @@
...
+    var textPreset4: FontWeightTypable { get }
+    var textPreset5: FontWeightTypable { get }
+    var textPreset6: FontWeightTypable { get }
+    var textPreset7: FontWeightTypable { get }
+    var textPreset8: FontWeightTypable { get }
+    var textPreset9: FontWeightTypable { get }
+    var textPreset10: FontWeightTypable { get }
+}
+
+public enum FontWeightTypable {
```
amegias, line 22:
```suggestion
/// Available font weights in Mistica
public enum MisticaFontWeightType {
```

Wdyt?? it's just a set of the allowed weights which will be mapped from the preset ones
https://github.com/Telefonica/mistica-ios/blob/df2b875d290ebf890a28435fb52dbb43755d1d5b/Sources/MisticaCommon/Fonts/FontStyle%2BWeights.swift#L27-L31

and finally mapped to the UIKit weights:
https://github.com/Telefonica/mistica-ios/blob/df2b875d290ebf890a28435fb52dbb43755d1d5b/Sources/MisticaCommon/Fonts/MisticaFontWeights.swift#L28-L39

amegias (approved):
Really good work!!!

cbarcom (approved):

AnaMontes11 (approved):
Nice job!

Sources/MisticaCommon/Fonts/MisticaFontWeights.swift
alejandroruizponce: Okey, tbh I wasn't sure with the naming, I like your idea. Fixed!

pbartolome:
I don't think most of these tests will need updates, are you sure you are running those using x86_64/Rosseta? also I see some images that have been updated with Spanish text (E.g Contactos instead of Contacts)

Sources/MisticaCommon/Fonts/FontStyle+Weights.swift
```diff
@@ -0,0 +1,99 @@
...
+            switch self {
+            case .light: return .light
+            case .regular: return .regular
+            case .medium: return .medium
+            }
+        }
+    }
+
+    enum TextPreset4Weight {
+        case light, regular, medium, branded
```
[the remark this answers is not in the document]
amegias: Talked with the Design Team, it will be finally a custom case for CardTitle label so it shouldn't be considered as a `.branded` textPreset4

amegias:
https://github.com/Telefonica/mistica-ios/blob/252092366a03bd3886417a1055853f4bd6caba90/MisticaCatalog/Source/FontsView.swift#L15-L20

now, it's not right at all :)
I would remove "bold"

Sources/MisticaCommon/Fonts/FontStyle+Weights.swift
```diff
@@ -0,0 +1,99 @@
...
+}
+
+struct VivoFontWeights: MisticaFontWeights {
+    public var cardTitlePreset: MisticaFontWeightType = .regular
+    public var textPreset5: MisticaFontWeightType = .regular
+    public var textPreset6: MisticaFontWeightType = .regular
+    public var textPreset7: MisticaFontWeightType = .regular
+    public var textPreset8: MisticaFontWeightType = .regular
+    public var textPreset9: MisticaFontWeightType = .regular
+    public var textPreset10: MisticaFontWeightType = .regular
```
amegias, line 98: I think you take the incoming new Vivo brand.

```suggestion
    public var textPreset5: MisticaFontWeightType = .light
    public var textPreset6: MisticaFontWeightType = .light
    public var textPreset7: MisticaFontWeightType = .light
    public var textPreset8: MisticaFontWeightType = .light
    public var textPreset9: MisticaFontWeightType = .light
    public var textPreset10: MisticaFontWeightType = .light
```

amegias (changes requested):

Sources/MisticaCommon/Fonts/FontStyle+Weights.swift
alejandroruizponce: Fixed!

amegias (approved):

alexanegon (approved):

pbartolome (approved):

PR #271, opened 2023-05-03
IOS-8817 Allow custom font weights

amegias:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-8817

## 🥅 **What's the goal?**
Allow customizing font weighs for custom mistica config like we already do for Colors and Assets. See how Zeus is configured:
https://github.com/Telefonica/zeus-ios/blob/5bc3f2963d0308a5d54f18f7f46e4cb68a4af0cd/App/iOS/AppDelegate.swift#L33

## 🚧 **How do we do it?**
Add a new associated value in `.custom` enum case.

Executed in Zeus:
Captura de pantalla 2023-05-03 a las 16 16 08

Before it was (MovES):
Captura de pantalla 2023-05-03 a las 16 17 07

### DIFF
--- a/Sources/MisticaCommon/Styles/BrandStyle.swift
+++ b/Sources/MisticaCommon/Styles/BrandStyle.swift
@@ -14,7 +14,7 @@ public enum BrandStyle {
     case vivo
     case o2
     case blau
-    case custom(MisticaColors, MisticaBrandAssets)
+    case custom(MisticaColors, MisticaBrandAssets, MisticaFontWeights)
 
     public var id: String {
         switch self {
--- a/MisticaCatalog/Source/ColorsView.swift
+++ b/MisticaCatalog/Source/ColorsView.swift
@@ -50,7 +50,7 @@ struct ColorsView: View {
             return O2ColorPalette()
         case .vivo:
             return VivoColorPalette()
-        case .custom(let colors, _):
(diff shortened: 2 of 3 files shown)


idenjoe (approved):

Sources/MisticaCommon/Styles/BrandStyle.swift
```diff
@@ -14,7 +14,7 @@ public enum BrandStyle {
     case vivo
     case o2
     case blau
-    case custom(MisticaColors, MisticaBrandAssets)
+    case custom(MisticaColors, MisticaBrandAssets, MisticaFontWeights)
```
alejandroruizponce, line 17: Note that this only allows you to customise from textPreset5 to textPreset10. But textPreset1 to textPreset4 would default to what is defined in Mistica. Just to point that out.

alejandroruizponce (approved):

Sources/MisticaCommon/Styles/BrandStyle.swift
amegias: yeah! you're right. It's the set that can be configured by OB. In the future, if we need to customize the textPreset10 (for example), we will add it to the protocol and MovES, Vivo,... **Solar** should be implement it ;)
Thanks

PR #273, opened 2023-05-04
New custom feedback type

pbartolome:
## 🎟️ **Jira ticket**
[IOS-8360](https://jira.tid.es/browse/IOS-8360)

## 🥅 **What's the goal?**
Design has a fourth type of feedback which is not implemented on iOS. This `feedback` type, is similar to the other ones, but support any custom icon:

MicrosoftTeams-image (2)

## 🚧 **How do we do it?**
- Create a new enum case `.feedback(icon)` to define this new type of feedback. As the image is something that only this new type can change, I've added as part of the enum case instead of adding it in the `FeedbackConfiguration` struct.
- Add tests
- Update doc


Sources/Mistica/Components/Feedback/README.md
```diff
@@ -13,7 +13,7 @@ Multiline strings that will be positioned one below the other. If the content is
 
 ### Style
 
-There are three styles available: `.success`, `.informative` and `.error`:
+There are three fixed styles available: `.success`, `.informative`, `.error`. There's one custom style `.feedback(image)` that can be used to provide an icon that will be used instead of the fixed ones:
 
 | Success     | Informative     | Error     |
```
amegias, line 18: 1. what about adding an example of Feedback?
2. Check the explanation below if it needs to be changed
> All styles, if presented modally, will show a 'close' button. If they are presented with a push, a back button will show up except when the style is `success`. In that case the actions should dismiss this screen.

Sources/MisticaSwiftUI/Components/Feedback/README.md
```diff
@@ -13,7 +13,7 @@ Multiline strings that will be positioned one below the other. If the content is
 
 ### Style
 
-There are three styles available: `.success`, `.informative` and `.error`:
+There are three fixed styles available: `.success`, `.informative`, `.error`. There's one custom style `.feedback(image)` that can be used to provide an icon that will be used instead of the fixed ones:
 
 | Success     | Informative     | Error     |
```
amegias, line 18: see my comment above

amegias (changes requested):

Sources/Mistica/Components/Feedback/FeedbackStyle+Toolkit.swift
```diff
@@ -75,30 +78,34 @@ extension FeedbackStyle {
...
+    var hapticFeedbackStyle: UINotificationFeedbackGenerator.FeedbackType? {
         switch self {
         case .success:
             return .success
         case .informative:
             return .warning
         case .error:
             return .error
+        case .feedback:
+            return nil
```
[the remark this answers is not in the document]
pbartolome: I have asked design about this, I would open another question in the public channel

amegias (approved):
Maybe we should evaluate to use only the SwiftUI component having a bridge (UIHostingVC) which returns a UIKit version:
- Remove duplicated doc
- Remove duplicated tests (screenshots -> size)
Anyway, out of scope.

Good work, Pablo!

alexanegon (approved):

PR #276, opened 2023-05-22
fix(HighlightedCard): actionButton duplicated for accessibility tool

alexanegon:
## 🎟️ **Jira ticket**
[IOS-8932](https://jira.tid.es/browse/IOS-8932)

## 🥅 **What's the goal?**
QA team reported that the button "Try to ask aura something" that corresponds to the `actionButton` of the `HighlightedCard` is duplicated in their accessibility tool.

## 🚧 **How do we do it?**
Remove it from the `accessibilityElements` because the button is accessible by default and now we do not have the `UIAccessibilityElement` for the card, removed in the [PR](https://github.com/Telefonica/mistica-ios/pull/275)

### DIFF
--- a/Sources/Mistica/Components/Cards/HighlightedCard.swift
+++ b/Sources/Mistica/Components/Cards/HighlightedCard.swift
@@ -143,7 +143,6 @@ public class HighlightedCard: UIView {
         get {
             [
                 horizontalStackView,
-                actionButton.isHidden ? nil : actionButton,
                 closeButton.isHidden ? nil : closeButton
             ].compactMap { $0 }
         }


Sources/Mistica/Components/Cards/HighlightedCard.swift
```diff
@@ -143,7 +143,6 @@ public class HighlightedCard: UIView {
         get {
             [
                 horizontalStackView,
-                actionButton.isHidden ? nil : actionButton,
```
salavert, line 146: I don't understand this change and the title of the PR

alexanegon: could you check it now? @salavert

amegias (approved):
Thanks for the explanation!

PR #287, opened 2023-06-27
IOS-9102 Fix corner radius bug when the corner radius is calculated b…

idenjoe:
## 🎟️ **Jira ticket**
[IOS-9102 ](https://jira.tid.es/browse/IOS-9102 ) Mistica CornerRadius not displayed correctly in UIKit in some cases

## 🥅 **What's the goal?**
The corner radius in some cases is calculated based on the view frame. So depending when you call this method the frame could be zero and you will get a 0 corner radius.

## 🚧 **How do we do it?**
I've changed from where that method is being called from (usually from the init method) to the layoutSublayers method to ensure that the view has a frame.

**There is an error when creating the snapshots of the buttons that I have not been able to solve at the moment. It only happens when the snapshot is generated. In the catalog the buttons are shown correctly. I'll create another ticket to solve this.
testRegularSizeWithPrimaryInverseStyle with-vivoNew-style

This could be checked in the Mistica Catalog from the ViVoNew brand.

## 🏑 **AppCenter build**

qr


yceballost (approved):
👌

pbartolome (approved):

alejandroruizponce (approved):
Great!

PR #288, opened 2023-07-10
Ios 9162 update token weights for native components

idenjoe:
## 🎟️ **Jira ticket**
[IOS-9162](https://jira.tid.es/browse/IOS-9162) Update token weights for components still using medium weight.

## 🥅 **What's the goal?**
Change all the existing components still using the medium weight with new tokens.

## 🚧 **How do we do it?**

1. Changed NovumBarButtonItem badgeTextAttributes from medium to indicator token.
2. Changed CroutonConfig croutonOverriddenSizes to use link token instead medium weight.
3. Changed UINavigationBar+Mistica misticaTitleTextAttributes to use navigationBar token instead of medium weight.
4. Changed title and description labels in LoadErrorViewController to use cardTitle token and regular weight instead of medium weight.
5. Changed MisticaAppearance normalTextAttributes and selectedTextAttributes to use tabsLabel token instead of medium weight.
6. Changed NavigationVew+Styleguide misticaTitleTextAttributes to use navigationBar token instead of medium weight.
7. Changed some tests snapshots.

You can use the following alpha.

## 🏑 **AppCenter build**

qr


alejandroruizponce (approved):
Great! 👍

Sources/Mistica/Components/ViewStates/LoadErrorViewController.swift
```diff
@@ -82,15 +82,15 @@ private extension LoadErrorViewController {
     }
 
     func setUpTitleText() {
-        titleTextLabel.font = .textPreset4(weight: .medium)
+        titleTextLabel.font = .textPreset4(weight: .cardTitle)
```
amegias, line 85: having `cardTitle` in ErrorViewController is weird @aweell . Could we avoid coupling the weight names with the components? Could we rename this?? (not needed right now)

amegias (approved):

Sources/Mistica/Components/ViewStates/LoadErrorViewController.swift
aweell: We use it also in the emptyStateCard component, i'm thinking if it makes sense to have a generic one that covers all kind of feedbacks using a text-preset smaller than 5. If it's not needed right now can we see if other new cases fall into the same category and then change it?

aweell (approved):

Sources/Mistica/Components/ViewStates/LoadErrorViewController.swift
amegias: yeah, not needed now but take in account to improve this ;)

PR #289, opened 2023-07-13
fix(InputField): Avoid TextField auto expands horizontally

WanaldinoTelefonica:
* IOS-8333

## 🎟️ **Jira ticket**
[IOS-8333](https://jira.tid.es/browse/IOS-8333)

## 🥅 **What's the goal?**
- Avoid TextField auto expands horizontally 
- Limit placeholder to one line

## 🚧 **How do we do it?**
- Set that the UITextField wrapped in the pomponent has low priority to resist being smaller, so when the text written is longer than the width it doen't expand.
- Set placeholder to one line, so if you set a long text the component do not deforms.

testLongAssistiveText 1
testLongPlaceholder 1
testLongText 1


alexanegon (approved):
Good job & congratulations on your first PR!

idenjoe (approved):
Good job!

PR #290, opened 2023-07-17
fix(Bottom Sheet): Fix component to expand horizontally

WanaldinoTelefonica:
## 🎟️ **Jira ticket**
[IOS-9113](https://jira.tid.es/browse/IOS-9113)

## 🥅 **What's the goal?**
- Fix the component to fit in iOS 14.5 or less. The texts are not well shown

## 🚧 **How do we do it?**
Info: The dot is centered on a single line label with the same specs than the visible label, but is in a deeper layer stackView.
- The component was using a dummy view to create the space for the dot in the front stackView. This dummy view has been removed and the stackView modifies the layout margin to replace this space.

Simulator Screenshot - iPhone 6s - 2023-07-17 at 12 13 34
Simulator Screenshot - iPhone 6s - 2023-07-17 at 12 14 51
Simulator Screenshot - iPhone 6s - 2023-07-17 at 12 14 56

### DIFF
--- a/Sources/Mistica/Components/Sheet/View/Fragments/List/Rows/InformativeRow.swift
+++ b/Sources/Mistica/Components/Sheet/View/Fragments/List/Rows/InformativeRow.swift
@@ -53,7 +53,7 @@ class InformativeRow: UIView {
 
     private let bottomContent: UIStackView = {
         let stackView = UIStackView()
-        stackView.spacing = 8
+        stackView.isLayoutMarginsRelativeArrangement = true
         return stackView
     }()
 
@@ -65,8 +65,6 @@ class InformativeRow: UIView {
         label.minHeight = 20
         return label
     }()
-
-    private lazy var dummyView = SpacerView(axis: .horizontal, amount: 1)
 }
 
 // MARK: Private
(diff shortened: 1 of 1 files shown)


Sources/Mistica/Components/Sheet/View/Fragments/List/Rows/InformativeRow.swift
```diff
@@ -84,13 +82,12 @@ private extension InformativeRow {
             bottomContent.isHidden = true
         }
 
-        dummyView.amount = item.icon.size.width
+        bottomContent.layoutMargins.left = item.icon.size.width + 8
```
alexanegon, line 85: IMHO Magic number "8" must be in a descriptive constant because you are using it here and in the `frontStackView` WDYT?

alexanegon (approved):

Gomay88 (approved):

Sources/Mistica/Components/Sheet/View/Fragments/List/Rows/InformativeRow.swift
WanaldinoTelefonica: Sure, I've changed it 😄

PR #293, opened 2023-07-25
feat(BottomSheet): Added SVG image format support.

WanaldinoTelefonica:
Added SDWebImageSVGNativeCoder dependency. Added SVG image format support.

## 🎟️ **Jira ticket**
[IOS-9147](https://jira.tid.es/browse/IOS-9147)

## 🥅 **What's the goal?**
- Add SVG image format support on BottomSheet

## 🚧 **How do we do it?**
- Add SDWebImage and SDWebImageSVGNativeCoder to download and decode images.

Some examples:
testSVGUrl with-vivo-style
testSVGDataUri with-blau-dark-style


Sources/Mistica/Utils/Extensions/UIImageView+URL.swift
```diff
@@ -26,11 +27,11 @@ extension UIImageView {
             self?.downloadImage(
```
amegias, line 27: why don't we use directly `SDWebImage`? It would simplify the code, wouldn't it?

WanaldinoTelefonica: I did a previous version where `SDWebImage` downloads the image and sets but tests didn't show the image. I decided to reverse to how it worked and only implement a decoder for SVG data to UIImage. The decoder is the same that the library explains to add as a coder (from the same project repo).

Captura de pantalla 2023-07-25 a las 13 15 26

If you consider to change into the full usage of `SDWebImage` I'll change it

amegias: I'm not an expert with that SDK but I would do it since the library recommends to use the main tool configuring it with the right decoder. Probably it has other interesting features like caching, etc.
The tests should work and if you have to register the decoder every time, you can remove+add decoder just to do this action idempotent if it already existed

alexanegon (commented):
I have some doubts:

1. What happens if the URL is not available? IMHO we must allow to have a placeholder image
2. It seems that [SVGNativeCoder](https://github.com/SDWebImage/SDWebImageSVGNativeCoder) does not support vector image.
3. Does this part of the library cache the image somehow or do you have to download it whenever you want?
4. Maybe we can use  [SVGURLImageGateway.swift](https://github.com/Telefonica/b2p-germany-ios/blob/main/Sources/Platform/Utils/SVGURLImageGateway.swift) for SwiftUI and create another one for UIKit

Sources/Mistica/Utils/Extensions/UIImageView+URL.swift
alexanegon: I am agree with @amegias I would use SDWebImage, you can review [SVGURLImageGateway.swift](https://github.com/Telefonica/b2p-germany-ios/blob/main/Sources/Platform/Utils/SVGURLImageGateway.swift)

amegias:
> I have some doubts:
> 
> 1. What happens if the URL is not available? IMHO we must allow to have a placeholder image

It will be included in other task

> 4. Maybe we can use  [SVGURLImageGateway.swift](https://github.com/Telefonica/b2p-germany-ios/blob/main/Sources/Platform/Utils/SVGURLImageGateway.swift) for SwiftUI and create another one for UIKit

Sadly Mistica cannot depend on novum-platform-ios, anyway we don't need all of those features :)

Good points!

WanaldinoTelefonica:
> I have some doubts:
> 
> 2. It seems that [SVGNativeCoder](https://github.com/SDWebImage/SDWebImageSVGNativeCoder) does not support vector image.

The [SVGNativeCoder](https://github.com/SDWebImage/SDWebImageSVGNativeCoder) one is defined as `Best and W3C standard` but as you said it doesn't support vector. On the other hand [SVGCoder](https://github.com/SDWebImage/SDWebImageSVGCoder) supports vector and bitmap but looks like is some buggy with some SVGs. Maybe we can add both dependencies as decoders but I don't think is th ebest option

WanaldinoTelefonica:
Tests cannot show images because the library dispatches the load/download of the images so the thread of the screenshot ends before the image is loaded.

amegias:
> > I have some doubts:
> > 
> > 2. It seems that [SVGNativeCoder](https://github.com/SDWebImage/SDWebImageSVGNativeCoder) does not support vector image.
> 
> The [SVGNativeCoder](https://github.com/SDWebImage/SDWebImageSVGNativeCoder) one is defined as `Best and W3C standard` but as you said it doesn't support vector. On the other hand [SVGCoder](https://github.com/SDWebImage/SDWebImageSVGCoder) supports vector and bitmap but looks like is some buggy with some SVGs. Maybe we can add both dependencies as decoders but I don't think is the best option

I would add just one. The developer doesn't want to choose a decoder xD

Sources/Mistica/Utils/Extensions/UIImageView+URL.swift
```diff
@@ -6,71 +6,24 @@
...
-                    }
+        let coder = SDImageSVGNativeCoder.shared
+        SDImageCodersManager.shared.removeCoder(coder)
+        SDImageCodersManager.shared.addCoder(coder)
+
+        sd_setImage(with: url) { [weak self] lightImage, _, _, _ in
+            self?.sd_setImage(with: urlForDarkMode) { darkImage, _, _, _ in
+                if let lightImage, let darkImage {
+                    lightImage.imageAsset?.register(darkImage, with: UITraitCollection(userInterfaceStyle: .dark))
+                    self?.image = lightImage
```
amegias, line 24: 🚨  If darkImage is nil, the light one is not set 🚨

```swift
        sd_setImage(with: url) { [weak self] lightImage, _, _, _ in
            self?.image = lightImage
            self?.sd_setImage(with: urlForDarkMode) { darkImage, _, _, _ in
                guard let darkImage else { return }
                lightImage?.imageAsset?.register(darkImage, with: UITraitCollection(userInterfaceStyle: .dark))
                self?.image = lightImage // I think it's not needed but we should test it
            }
        }
```

WanaldinoTelefonica: That's right, but the first `self?.image = lightImage` is not needed as `sd_setImage` already does that. I'm fixing it now

amegias: thanks!!

amegias (approved):

Sources/Mistica/Utils/Extensions/UIImageView+URL.swift
```diff
@@ -6,71 +6,27 @@
...
-                urlForDarkMode: urlForDarkMode
-            ) { [weak self] lightImageData, darkImageData in
-                guard let lightImageData else { return }
-                self?.updateImage(imageData: lightImageData, darkImageData: darkImageData)
-            }
-            return
-        }
+        let coder = SDImageSVGNativeCoder.shared
+        SDImageCodersManager.shared.removeCoder(coder)
+        SDImageCodersManager.shared.addCoder(coder)
```
alexanegon, line 18: Why do you need to remove the coder? Could we check if the coder is inside the array or is the cost of the operation bigger?

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogSheetViewController.swift
```diff
@@ -232,6 +245,24 @@ private extension UICatalogSheetViewController {
...
...DUwLjIsMTYuNSw1MS4zeiIvPgoJPHBhdGggZD0iTTksMzEuNmMzLjUsMy45LDcuMiw3LjYsMTEuMSwxMS4xYzAuOC0xLjYsMS43LTMuMSwyLjYtNC42YzAuMS0wLjIsMC4zLTAuNCwwLjQtMC42Yy0yLjktMy4zLTMuMS05LjItMC42LTE3LjYgICBjMC44LTIuNywxLjgtNS4zLDIuNy03LjRjLTUuMiwzLjQtOS44LDgtMTMuMywxMy43QzEwLjgsMjcuOSw5LjgsMjkuNyw5LDMxLjZ6Ii8+Cgk8cGF0aCBkPSJNMTUuNCw1NC43Yy0yLjYtMS02LjEsMC43LTkuNywzLjRjMS4yLDYuNiwzLjksMTMsOCwxOC41QzEzLDY5LjMsMTMuNSw2MS44LDE1LjQsNTQuN3oiLz4KCTxwYXRoIGQ9Ik0zOS44LDU3LjZDNTQuMyw2Ni43LDcwLDczLDg2LjUsNzYuNGMwLjYtMC44LDEuMS0xLjYsMS43LTIuNWM0LjgtNy43LDctMTYuMyw2LjgtMjQuOGMtMTMuOC05LjMtMzEuMy04LjQtNDUuOC03LjcgICBjLTkuNSwwLjUtMTcuOCwwLjktMjMuMi0xLjdjLTAuMSwwLjEtMC4yLDAuMy0wLjMsMC40Yy0xLDEuNy0yLDMuNC0yLjksNS4xQzI4LjIsNDkuNywzMy44LDUzLjksMzkuOCw1Ny42eiIvPgoJPHBhdGggZD0iTTI2LjIsODguMmMzLjMsMiw2LjcsMy42LDEwLjIsNC43Yy0zLjUtNi4yLTYuMy0xMi42LTguOC0xOC41Yy0zLjEtNy4yLTUuOC0xMy41LTktMTcuMmMtMS45LDgtMiwxNi40LTAuMywyNC43ICAgQzIwLjYsODQuMiwyMy4yLDg2LjMsMjYuMiw4OC4yeiIvPgoJPHBhdGggZD0iTTMwLjksNzNjMi45LDYuOCw2LjEsMTQuNCwxMC41LDIxLjJjMTUuNiwzLDMyLTIuMyw0Mi42LTE0LjZDNjcuNyw3Niw1Mi4yLDY5LjYsMzcuOSw2MC43QzMyLDU3LDI2LjUsNTMsMjEuMyw0OC42ICAgYy0wLjYsMS41LTEuMiwzLTEuNyw0LjZDMjQuMSw1Ny4xLDI3LjMsNjQuNSwzMC45LDczeiIvPgo8L2c+Cjwvc3ZnPg=="
```
alexanegon, line 252: IMHO I would add this "big" string to a file and extract it in the code, but it is just my feeling if you are comfortable go on :)

Sources/Mistica/Utils/Extensions/UIImageView+URL.swift
amegias: IMO it's simpler than adding conditionals like:
```swift
if SDImageCodersManager.shared.coders.filter { $0 == coder } == nil {
       SDImageCodersManager.shared.addCoder(coder)
}
```

alexanegon (approved):

PR #294, opened 2023-08-08
Ios 9080 button chevron

WanaldinoTelefonica:
## 🎟️ **Jira ticket**
[IOS-9080](https://jira.tid.es/browse/IOS-9080)

## 🥅 **What's the goal?**
- Add a chevron image for linkButton

## 🚧 **How do we do it?**
- For UIKit define a new property that configures an UIImageView with the corresponding image.
- There is an enum with the posible cases with only chevron defined. This could be extended in the future.

- For SwiftUI there is a new property `withChevron` on the `ButtonStyle` for `misticaLink` and `misticaLinkInverse`. This will add on `MisticaButtonStyle` the property `.chevron` on `MisticaButton` for `rightImage`

Check screenshots :D

## 🏑 **AppCenter build**
Use last version 25.4.0 (1)
https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-swiftui-ios/distribution_groups/public/releases/22


yceballost:
please add @mistica-design as a reviewer :)

amegias:
When the link isLoading, is the chevron hidden?

amegias (approved):
good work!

yceballost:
Can you generate a mistica catalog build to review it? I think you can deploy it with [an action ](https://github.com/Telefonica/mistica-ios/actions/workflows/build-mistica-catalog.yml) (confirm with @amegias)

Sources/Mistica/Components/Button/Button.swift
```diff
@@ -56,6 +63,29 @@ open class Button: UIControl {
...
+        private var _image: UIImage {
+            switch self {
+            case .chevron: return .chevron
+            case let .custom(image: image): return image
+            }
+        }
+
+        var image: UIImage {
+            _image.withRenderingMode(.alwaysTemplate)
+        }
```
DavidMarinCalleja, line 79: Why have `image: UIImage` and `_image: UIImage`?

```suggestion
        var image: UIImage {
            switch self {        
            case .chevron: 
                return .chevron.withRenderingMode(.alwaysTemplate)
            case let .custom(image: image):
                return image.withRenderingMode(.alwaysTemplate)
            }                               
        }
```

Sources/Mistica/Components/Button/ButtonContentView.swift
```diff
@@ -10,13 +10,31 @@ import UIKit
...
     }()
 
+    private let rightImageView: UIImageView = {
+        let imageView = UIImageView()
+        imageView.contentMode = .scaleAspectFit
+        imageView.isHidden = true
+        return imageView
+    }()
+
+    private lazy var rightImageHeightConstraint: NSLayoutConstraint = rightImageView.heightAnchor.constraint(equalToConstant: 1)
```
DavidMarinCalleja, line 36: because the value of the constant is 1

Sources/Mistica/Components/Button/Button.swift
WanaldinoTelefonica: The point is that the image should have the same color as the title, so it should be rendered as a template to be tinted. The idea is that I didn't want to set all the image cases with `.withRenderingMode(.alwaysTemplate)`. The private `_image` is the image itself and the published property `image` is which sets the property

Sources/Mistica/Components/Button/ButtonContentView.swift
WanaldinoTelefonica: It is just an arbitrary value for the definition of the constraint

WanaldinoTelefonica:
> When the link isLoading, is the chevron hidden?

Yes, AFAIK the chevron only shown on the right of the main title, not on loading

aweell (commented):
I see the right and the left alignment, in order to anticipate future evolutions, does it make sense to add the vertical bleeding alignment in this PR? When the buttonLink goes into a title, the size of the container is making the title container to be bigger than it should be, that's why the need of this property. https://jira.tid.es/browse/IOS-9201

aweell (commented):
When I change the preferences of font size in the device in UIKit (In SwiftUI works fine), the primary, secondary and danger buttons seem to react increasing the font size, but the buttonLink stays always unaltered, is this an already known issue?

WanaldinoTelefonica:
> I see the right and the left alignment, in order to anticipate future evolutions, does it make sense to add the vertical bleeding alignment in this PR? When the buttonLink goes into a title, the size of the button container is making the title container to be bigger than it should be, that's why the need of this property. https://jira.tid.es/browse/IOS-9201

I think this should only target adding the chevron for creating the new bottomsheet functionality



> When I change the preferences of font size in the device in UIKit (In SwiftUI works fine), the primary, secondary and danger buttons seem to react increasing the font size, but the buttonLink stays always unaltered, is this an already known issue?

Yes, we know but it was deprioritized. If there is time before the release we have this task to add it https://jira.tid.es/browse/IOS-9244

aweell (approved):
LGTM!

DavidMarinCalleja (approved):

PR #297, opened 2023-08-16
LATCH-918 Fix color for filling and borders in Checkbox

jaranaz:
## 🎟️ **Jira ticket**

https://jira.tid.es/browse/LATCH-918

## 🥅 **What's the goal?**

In last release 25.4.0, borders and filling color for the Checkbox component are wrong.
Fix the component to restore its look & feel.

## 🚧 **How do we do it?**

- Define new property colors `fillColor` and `borderColor`
- Border line width is always 1
- Use the modifier `fill()` and `border()` to set the right colors

Screenshot 2023-08-17 at 10 13 51

261035898-2310b5e8-3293-4a1c-aa87-4b7fddfc5978

261035928-76ab177d-f00e-48a8-824c-eedddc1f9186


WanaldinoTelefonica (approved):
🌻

Sources/MisticaSwiftUI/Components/Checkbox/Checkbox.swift
```diff
@@ -28,12 +29,12 @@ public struct Checkbox: View {
             }
     }
 
-    var controlColor: Color {
-        isSelected ? .controlActivated : .control
+    var fillColor: Color {
```
idenjoe, line 32: I'm not sure if clear color fits for all brands.

Sources/MisticaSwiftUI/Components/Checkbox/Checkbox.swift
```diff
@@ -28,12 +29,12 @@ public struct Checkbox: View {
...
+        isSelected ? .controlActivated : .clear
```
idenjoe, line 33: I'm not sure if clear color fits for all brands.

jaranaz: I would think that the effect was the same than before, when we were using RoundedRectangle() instead of Rectangle(). The fill was clear colored, but without setting clear color explicitly 🤔

Let's check with @yceballost! 🙂

yceballost: Can you include a dark mode screenshot to check this? 

The result should be the same as web
https://tinyurl.com/264tr4qq

yceballost: We have to use `background` color in this case 

Web implementation:
https://github.com/Telefonica/mistica-web/blob/73f91ac107127b545021be68f4e498e940ab192f/src/checkbox.css.ts#L35C32-L35C44

(In Figma we also use `background`)

jaranaz: @idenjoe @yceballost 
Done! changed from .clear to .background color

WanaldinoTelefonica (approved):

yceballost (approved):

idenjoe (approved):

salavert (approved):

PR #303, opened 2023-09-12
iOS-8994 feed back evolution

WanaldinoTelefonica:
## 🎟️ **Jira ticket**
[IOS-8994](https://jira.tid.es/browse/IOS-8994)

## 🥅 **What's the goal?**
- Reduce side margins from 24px to 16px
- Decrease icon size from 64 to 48
- Custom success animation for Vivo_New
- Add the new way to show the info animated

## 🚧 **How do we do it?**
- Change properties
- Replace json for new animation
- Rewrite how the info is shown in view

https://github.com/Telefonica/mistica-ios/assets/139126842/b33f6bd2-8ebe-4345-9c99-74940198cd36

## 🏑 **AppCenter build**
image


aweell (commented):
Screenshot 2023-09-12 at 11 52 34

It seems that the asset in the error and success feedbacks is 48x48 but the info and feedback styles are keeping the 64 size?

Also the content of the latter is not animated?

WanaldinoTelefonica:
> Screenshot 2023-09-12 at 11 52 34
> 
> It seems that the asset in the error and success feedbacks is 48x48 but the info and feedback styles are keeping the 64 size?
> 
> Also the content of the latter is not animated?

Yes, thats right! Non animated images didn't resize into 48px correctly. Thanks!  
New version generated with the fix 😄

aweell:
> What do you mean that latter content is not animated? If you refer to info style, specs only define that succes and error should.

Sorry, not explicit enough with "latter" I talking about the feedback style variant. The [current web implementation](https://mistica-web.vercel.app/?path=/story/patterns-feedback-feedbackscreen--feedback-screen-story) was showing the animation but is because is being used as a base component for the others and the animation is configurable via prop. Forget about this :)

Sources/Mistica/Components/Feedback/FeedbackView.swift
```diff
@@ -290,32 +322,86 @@ private extension FeedbackView {
...
+        var views = [UIView]()
+        if title.isEmpty == false {
+            views.append(titleLabel)
+        }
+        if let subtitle, subtitle.isEmpty == false {
+            views.append(subtitleLabel)
+        }
+        if let errorReference, errorReference.isEmpty == false {
+            views.append(errorReferenceLabel)
+        }
```
DavidMarinCalleja, line 362: This can be simplified by using the `!` operator
```suggestion
        if !title.isEmpty {
            views.append(titleLabel)
        }
        if let subtitle, !subtitle.isEmpty {
            views.append(subtitleLabel)
        }
        if let errorReference, !errorReference.isEmpty {
            views.append(errorReferenceLabel)
        }
```

Tests/MisticaTests/UI/__Snapshots__/FeedbackTests/testSuccessTitle.with-vivoNew-dark-style.png
DavidMarinCalleja, line 1: In this case, Is this going to change the Viviño icon to Tick icon?

WanaldinoTelefonica: Yes, only for success feedback screen

Sources/Mistica/Components/Feedback/FeedbackView.swift
WanaldinoTelefonica: I'm not a big fan of inverting booleans as I consider it is less readable.

DavidMarinCalleja (approved):

idenjoe (approved):

WanaldinoTelefonica:
Hi @aweell! A new version has been upload with the new json file. Thanks 😄

aweell (approved):
Nice job!

PR #304, opened 2023-09-22
LATCH-987 Fix Snackbar button color

jaranaz:
## 🎟️ **Jira ticket**

https://jira.tid.es/browse/LATCH-987

## 🥅 **What's the goal?**

Component `Snackbar` is using token `textLink` for its action button, instead of the right token `textLinkSnackbar`

## 🚧 **How do we do it?**

Create new `MisticaButtonStyle` type called `.misticaSnackbar` with the right color token.

Screenshot 2023-09-22 at 12 06 50

Screenshot 2023-09-22 at 12 07 01


pbartolome (approved):

yceballost (approved):

Sources/MisticaSwiftUI/Components/Button/MisticaButtonStyle.swift
```diff
@@ -271,4 +271,41 @@ public extension ButtonStyle where Self == MisticaButtonStyle {
             rightImage: withChevron ? .chevron : nil
         )
     }
+
+    static func misticaSnackbar(
+        small: Bool = false,
+        bleedingAlignment: ButtonBleedingAlignment = .none,
+        withChevron: Bool = false
```
WanaldinoTelefonica, line 278: IMO this should not be avalable as chevron is only available on link style

WanaldinoTelefonica (approved):
👍

PR #305, opened 2023-09-22
Ios 9244 button link accesibility

alexanegon:
## 🎟️ **Jira ticket**
[IOS-9244](https://jira.tid.es/browse/IOS-9244)

## 🥅 **What's the goal?**
We must enable that the right image of the `MisticaButton` follows the accesibility property

## 🚧 **How do we do it?**
We have a new property wrapper in iOS 14 called `ScaledMetric`  it is to define numbers that should scale automatically according to the user’s Dynamic Type settings.

To use it due to support iOS 13 we have to create a new struct not available prior to iOS 14 called `ScaledMisticaButton` which receives a content, height and width in its init to keep the functionality of the old modifiers and extend them with the scale

https://github.com/Telefonica/mistica-ios/assets/43632903/2b02ffa0-45bb-45af-8626-37bcfa0cc08e


Sources/MisticaSwiftUI/Components/Button/ScaledMisticaButton.swift
```diff
@@ -0,0 +1,65 @@
...
+        content()
+            .frame(height: height * scale)
+            .if(!small) { $0.expandHorizontally() }
+            .frame(minWidth: minWidth * scale)
+            .padding(.vertical, verticalPadding)
+            .padding(.horizontal, horizontalPadding)
+            .background(backgroundColor)
+            .foregroundColor(foregroundColor)
+            .border(radiusStyle: .button, borderColor: borderColor)
+            .padding(EdgeInsets(top: 0, leading: leadingInset, bottom: 0, trailing: trailingInset))
```
amegias, line 63: those lines don't depend on the `scale`, could we put in common with the other approach?

amegias (approved):

WanaldinoTelefonica (approved):

Sources/MisticaSwiftUI/Components/Button/ScaledMisticaButton.swift
alexanegon: That was the my first approach but they modify the button layout and breaks the tests :(

PR #308, opened 2023-10-04
LATCH-1000: fix Button disabled state

jaranaz:
## 🎟️ **Jira ticket**

https://jira.tid.es/browse/LATCH-1000

## 🥅 **What's the goal?**

The `Button` component for SwiftUI is using wrong colors for its disabled state.
According to its equivalent in UIKit, it should apply and alpha of 0.5 to its background color.

## 🚧 **How do we do it?**

Set background color with alpha 0.5 when disabled.

### Before ###

before

### After ###
after


dhidalgofadrique (approved):
👍

WanaldinoTelefonica (approved):
🚀

PR #309, opened 2023-10-04
LATCH-1002: update text and background colors for Tag

jaranaz:
## 🎟️ **Jira ticket**

https://jira.tid.es/browse/LATCH-1002

## 🥅 **What's the goal?**

The `Tag` component is not using the right text and background colors for its styles, according to [the specs](https://www.figma.com/file/wseConngNv2MYEXgCk5uI4/%F0%9F%94%B8-Tag-Specs?type=design&node-id=1530-1344&mode=design&t=YZ85aqAC9BVbb9YE-4) for this component. 

## 🚧 **How do we do it?**

Fix `Tag` component colors in both UIKit and SwiftUI.

## 🏑 **AppCenter build**

qrcode (1)


Sources/MisticaSwiftUI/Components/Tag/Tag.swift
```diff
@@ -104,7 +104,7 @@ public struct Tag: View {
         case .inactive:
             return .neutralMedium
         case .warning:
-            return .warning
+            return .warningHigh
```
idenjoe, line 107: Maybe we should check all the styles and their corresponding token. In TagViewStyle we have different tokens for promo, active, success, error styles (and inverse that does not exist in SwiftUI?)

yceballost (approved):

alexanegon (approved):

Sources/MisticaSwiftUI/Components/Tag/Tag.swift
jaranaz: I'm going to check on this, probably we will need to fix more than warning style

jaranaz: Done! Now we have fixed the wrong colors, according to the specs document ;)

alexanegon (approved):

idenjoe (approved):

PR #312, opened 2023-10-13
IOS-9261 Add public titleTinColor property to be able to change the t…

idenjoe:
## 🎟️ **Jira ticket**
[IOS-9261](https://jira.tid.es/browse/IOS-9261) Add titleTintColor public property to be able to set the title color from the clients.

## 🥅 **What's the goal?**
Due a requirement for Vivo digital hub experience we need to set the title color for the center section.

## 🚧 **How do we do it?**
Added a new public property to be able to set the tint color to the title.


yceballost:
Hi guys, this is not allowed by rowList definition in specs :(. You can't open a PR evolving a component without take it in account design specs or mística design team

idenjoe:
> Hi guys, this is not allowed by rowList definition in specs :(. You can't open a PR evolving a component without take it in account design specs or mística design team

@yceballost as we commented a few days ago we finally will allow this change but restricting allowed Colors. So now you only can set normal or danger titleTextColor that will match with textPrimary or textLinkDanger.

Sources/Mistica/Components/Lists/ListCellContentView.swift
```diff
@@ -84,6 +98,12 @@ open class ListCellContentView: UIView {
         }
     }
 
+    public var titleTextColorType: TitleTextColorType = .normal {
+        didSet {
+            centerSection.titleTextColor = titleTextColorType.value
```
DavidMarinCalleja, line 103: During setup, the `titleTextColor` is set to `cellStyle.titleTextColor` 
```
    func updateCellStyle() {
        centerSection.titleTextColor = cellStyle.titleTextColor
```

shouldn't this new call be used to initialise to the two possible values (normal or danger)?
```
    func updateCellStyle() {
        titleTextColorType = .normal
```

DavidMarinCalleja: talked offline and clarified with @idenjoe

yceballost (approved):

jaranaz (approved):

PR #315, opened 2023-10-17
[IOS-9386] Building with XCode 15

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9386

## 🥅 **What's the goal?**
Build and integrate with XCode 15 and iOS 17 now we have slaves updated with that version.

## 🚧 **How do we do it?**
Update configuration in Makefile with a new simulator iPhone 15 in iOS 17.0 and using XCode 15 app.

Compiling with XCode 15 and on iPhone 15 misaligns some tiny pixels in the snapshots which requires that many of them have been re-generated with little difference.

Check CI tests and new catalog version generated.


amegias:
@alejandroruizponce where is the app? Remember to upload it in the Alpha slot
https://appcenter.ms/orgs/Tuenti-Organization/apps/Mistica-SwiftUI-iOS

DavidMarinCalleja (approved):
Looks good to me, but I just have a question, do you think it is in our interest to add a tolerance margin to avoid uploading images that are actually almost identical?

amegias:
> Looks good to me, but I just have a question, do you think it is in our interest to add a tolerance margin to avoid uploading images that are actually almost identical?

It won't solve this kind of PRs :( since you record the test and the result is different (new screenshot needs to be commited :()
I would avoid testing for all brands in some cases which are not interesting

alejandroruizponce:
> @alejandroruizponce where is the app? Remember to upload it in the Alpha slot https://appcenter.ms/orgs/Tuenti-Organization/apps/Mistica-SwiftUI-iOS

Ok, now it's uploaded in alpha. I had uploaded it in enterprise instead.

amegias (approved):

PR #317, opened 2023-10-25
LATCH-1028 Fix DataCard according to specs

jaranaz:
## 🎟️ **Jira ticket**

https://jira.tid.es/browse/LATCH-1028

## 🥅 **What's the goal?**

According to the [specs](https://www.figma.com/file/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?type=design&node-id=201-6943&mode=design&t=GJLRKKaNGsmNZEXw-0) of DataCard component, it should let title or description empty, and the absence of primary and link buttons.

Thus, modify both components DataCard in UIKit and SwiftUI frameworks.

## 🚧 **How do we do it?**

- Set `title` and `description` as optionals
- New inits without params for `primaryButton` and `linkButton`
- Fix some paddings to follow [specs](https://www.figma.com/file/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?type=design&node-id=201-6943&mode=design&t=GJLRKKaNGsmNZEXw-0)
- Update snapshots
- Update Mistica app components to reflect these changes

## 🏑 **AppCenter build**

qrcode (4)


yceballost (approved):
lgtm! :)

Sources/MisticaSwiftUI/Components/Cards/DataCard.swift
```diff
@@ -93,41 +95,43 @@ public struct DataCard<Headline: View, Fragment: View, PrimaryButton: View, Link
...
                 fragmentView
                     .padding(.top, 16)
             }
 
             // This spacer along with `fixedVerticalContentSize = true` makes the content of the card to expand.
             Spacer()
 
-            HStack(spacing: 24) {
-                if let primaryButton = primaryButton {
+            if hasPrimaryButton || hasLinkButton {
```
idenjoe, line 117: What happens if hasPrimaryButton is true and hasLinkButton is false? linkButton is not optional...

idenjoe (approved):

Sources/MisticaSwiftUI/Components/Cards/DataCard.swift
jaranaz: In that case, only `primaryButton` would be drawn.
None of them, `primaryButton` or `linkButton`, could be nil, just EmptyView in case they are not used (that is why we use the methods `hasPrimaryButton` and `hasLinkButton`, to check if they are an EmptyView or not)

PR #318, opened 2023-10-27
[IOS-9461] Fix URL(string:) behavior

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9461

## 🥅 **What's the goal?**
Apple’s URL parsing implemented RFC1738/1808 standard and was changed to RFC3986 starting from iOS 17. Actually, it can produce problems with code written before, because now URL(string: "Invalid URL") is not nil. Apple updated the URL initializer for iOS 17 with a new Bool parameter encodingInvalidCharacters that has a default value (true) — it is what makes the behavior unpredictable and it needs to be handled.

## 🚧 **How do we do it?**
We have added a custom initializer where we will pass our urlString and behind if we are in iOS 17 we will set the new parameter encodingInvalidCharacters to false so that the behavior is similar to what we had before and return nil if the URL is not valid.

This custom initializer will be the one that by default we will have to use for new URLs that we want to validate and the one that replaces all the previous initializers that we already have implemented.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.9
 
 import PackageDescription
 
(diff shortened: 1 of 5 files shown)


dhidalgofadrique (approved):
👍

WanaldinoTelefonica (approved):
👍

PR #322, opened 2023-11-08
Ios 9391 snackbar infinite actions

idenjoe:
## 🎟️ **Jira ticket**
[IOS-9391](https://jira.tid.es/browse/IOS-9391) Snackbar Infinite time interval

## 🥅 **What's the goal?**
We need to support infinite intervals and an optional close button that could appear in some conditions.

## 🚧 **How do we do it?**
1. Add support for infinite CroutonDismissInterval
2. Add support for a close button:
- Forced
- In case the crouton has an infinite dismiss interval and no action
3. Added some logic to calculate the correct dismiss interval for some conditions. You can see these conditions [here](https://www.figma.com/file/yCHLIfy4WMfRdlADwyL4kZ/🔸-Snackbar-Specs?type=design&node-id=1038-1306&mode=design&t=W7B91J8NGCxOS4UO-0)

You can test it in Mistica Catalog.

## 🏑 **AppCenter build**
https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-swiftui-ios/distribution_groups/public/releases/46
qrcode_install appcenter ms


Sources/Mistica/Components/Crouton/Domain/CroutonConfig.swift
```diff
@@ -17,6 +17,8 @@ import UIKit
 public enum CroutonDismissInterval: Int, CaseIterable {
     case fiveSeconds = 5
     case tenSeconds = 10
+    // The infinite rawValue will/should not be used to dispatch a timer.
+    case infinite = 999
 }
```
amegias, line 22: I would do:
```suggestion
public enum CroutonDismissInterval: CaseIterable {
    case fiveSeconds
    case tenSeconds
    case infinite
}

public extension CroutonDismissInterval {
    var timeInterval: TimeInterval? {
        switch self {
            case .fiveSeconds:
                return 5
            case .tenSeconds:
                return 10
            case .infinite:
                return nil
        }
    }
}
```

and then
```diff
    func addCountdownToDismiss() {
-        guard config.overrideDismissInterval != .infinite else {
+        guard let timeInterval = config.overrideDismissInterval.timeInterval else {
            return
        }

        timer = Timer.scheduledTimer(
-            timeInterval: config.overrideDismissInterval.timeInterval,
+            timeInterval: timeInterval,
            target: self,
            selector: #selector(invokeDismissHandler),
            userInfo: nil,
            repeats: false
        )
    }
```

Sources/Mistica/Components/Crouton/Domain/CroutonConfig.swift
```diff
@@ -17,6 +17,8 @@ import UIKit
 public enum CroutonDismissInterval: Int, CaseIterable {
     case fiveSeconds = 5
     case tenSeconds = 10
+    // The infinite rawValue will/should not be used to dispatch a timer.
+    case infinite = 999
```
idenjoe, line 21: The int value will not be used to setup an interval. I will change this on Friday in another PR to have two different enums. One with something like timeInterval(CroutonDismissInterval) and inifinite and I will remove the infinite from this enum.

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
```diff
@@ -224,4 +229,23 @@ private extension CroutonController {
 
         return viewController
     }
+
+    /// Calculate the CroutonDismissInterval depending on the action and croutonDismissInterval.
+    /// - Parameters:
+    ///   - action: An optional action which will show a button with the given title and invoke the handler when the button is pressed
+    ///   - croutonDismissInterval: The time interval that the crouton should be displayed.
+    /// - Returns: This method will take into account the interval that the user wants and whether or not it has action to return a CroutonDisssmisInterval. In certain incompatible cases, the interval returned will be different from the one the user has selected.
+    func croutonDismissInterval(from action: ActionConfig?, croutonDismissInterval: CroutonDismissInterval?) -> CroutonDismissInterval {
```
idenjoe, line 238: We need to calculate the right dismiss interval because it could be configured from outside with incompatible values with the Snackbar [specs](https://www.figma.com/file/yCHLIfy4WMfRdlADwyL4kZ/🔸-Snackbar-Specs?type=design&node-id=1038-1306&mode=design&t=OBkiqJR0qsU4oj2y-0)

Sources/Mistica/Components/Crouton/Presentation/CroutonView.swift
```diff
@@ -61,12 +63,40 @@ class CroutonView: UIView {
         return stackView
     }()
 
-    private lazy var stackView: UIStackView = {
+    private lazy var verticalStackView: UIStackView = {
```
idenjoe, line 66: I've created two different stacks instead of changing the axis of the old one we had.

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
```diff
@@ -224,4 +229,23 @@ private extension CroutonController {
...
+
+        if action == nil && croutonDismissInterval == .tenSeconds {
+            return .fiveSeconds
+        } else if action != nil && croutonDismissInterval == .fiveSeconds {
+            return .tenSeconds
+        } else {
+            return dismissInterval
+        }
+    }
 }
```
amegias, line 251: I would do something like:

```suggestion
    func normalizeDismissInterval(_ croutonDismissInterval: CroutonDismissInterval?, action: ActionConfig?) -> CroutonDismissInterval {
        switch croutonDismissInterval {
            // Handled cases
        case .none where action != nil:
            return .tenSeconds
        case .fiveSeconds where action != nil:
            return .tenSeconds
        case .tenSeconds where action == nil:
            return .fiveSeconds
        case .none:
            return .fiveSeconds
            
            // Default:
        case .fiveSeconds:
            return .fiveSeconds
        case .tenSeconds:
            return .tenSeconds
        case .infinite:
            return .infinite
        }
    }
```

- I've renamed the method since its responsibility is to fix/normalize/sanitize the final value given an action and a raw crouton interval.
- Using `switch` we will be aware of adding a new case which will have to be handled here (or a warning will be thrown)
- Using `switch` looks more readable. Otherwise I would add unit tests here :/

Anyway, this logic is

amegias (approved):

amegias:
I've seen a warning coming from the new XCode:
https://github.com/Telefonica/mistica-ios/blob/b1218cc85e0b38798095ade8d7d2bfe7060dd426/Sources/MisticaSwiftUI/Components/Carousel/Carousel.swift#L52
Captura de pantalla 2023-11-08 a las 17 09 09
Could you fix it please? Just do:
```diff
-    func stack<Content: View>(@ViewBuilder content: () -> Content) -> some View {
+    func stack<ContentView: View>(@ViewBuilder content: () -> ContentView) -> some View {
```

Sources/Mistica/Components/Crouton/Domain/CroutonConfig.swift
idenjoe: Fixed: https://github.com/Telefonica/mistica-ios/pull/322/commits/7d25de85d99078554f37a048c699a4a9f9b0710b

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
idenjoe: Fixed:https://github.com/Telefonica/mistica-ios/pull/322/commits/7d25de85d99078554f37a048c699a4a9f9b0710b

DavidMarinCalleja (approved):

yceballost (approved):
lgtm!

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
```diff
@@ -8,12 +8,49 @@
 
 import UIKit
 
+@objc public enum CroutonControllerDismissReason: Int, RawRepresentable {
```
amegias, line 11: objc?

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
```diff
@@ -8,12 +8,49 @@
...
+    case dismiss
+    case button
+    case timeout
+    case consecutive
```
amegias, line 15: dead code. I wouldn't add it...

amegias (approved):

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
idenjoe: Ok. I remove the consecutive case for now.

DavidMarinCalleja: I would remove consecutive if it is no longer used

idenjoe: Done

PR #323, opened 2023-11-13
[IOS-9505] Set custom size in image rows

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9505

## 🥅 **What's the goal?**
Due to the new design requirements for the rows part of List defined here: https://www.figma.com/file/PApssATUkePZnIOB4GmAeI/%F0%9F%94%B9-%E2%80%8E%E2%80%8ETokens-definition?type=design&node-id=1549-274&mode=design&t=516M7DFAZXJKpCzl-0

It is considered necessary that a row image can be defined with a custom size or a custom aspect ratio. In this case, and for the ease for the developer it has been decided to implement a custom size, since it is the most intuitive and simple to implement.

## 🚧 **How do we do it?**
In MisticaSwiftUI this requirement is already implemented through .squaredImage:
https://github.com/Telefonica/mistica-ios/blob/14a168ec49ab432dab20bcd02453442965cbb628/Sources/MisticaSwiftUI/Components/List/Cell.swift#L28

For Mistica from UIKit we've refactorized the .image case:
https://github.com/Telefonica/mistica-ios/blob/14a168ec49ab432dab20bcd02453442965cbb628/Sources/Mistica/Components/Lists/ListCellContentView.swift#L49
It has been refactored to accept a nullable parameter of type CGSize. 
https://github.com/Telefonica/mistica-ios/blob/f434299c2180f260cec76ff2f0f6bf8d218a9803/Sources/Mistica/Components/Lists/ListCellContentView.swift#L49
In the case that this parameter is configured, its radius will also be configured according to the type of token for the image, which is the .container:
https://github.com/Telefonica/mistica-ios/blob/f434299c2180f260cec76ff2f0f6bf8d218a9803/Sources/Mistica/Components/Lists/Internals/CellLeftSectionView.swift#L120

One way would be to add a custom size in the Catalog of UIKit, specifically in its default parameter:
https://github.com/Telefonica/mistica-ios/blob/f434299c2180f260cec76ff2f0f6bf8d218a9803/MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogListsViewController.swift#L166


amegias (commented):
Thanks for the explanation! It looks great!! I miss:
1. changes in SwiftUI Cell
2. Add this configuration to our catalog to test it
Captura de pantalla 2023-11-13 a las 15 54 32
3. Add a couple of snapshot tests

Sources/Mistica/Components/Lists/ListCellContentView.swift
```diff
@@ -46,7 +46,7 @@ open class ListCellContentView: UIView {
     @frozen
     public enum CellAssetType: Equatable {
         case none
-        case image(UIImage)
+        case image(UIImage, size: CGSize? = nil)
```
amegias, line 49: not sure if we should allow any size or just force some aspect ratios like:
```suggestion
        case image(UIImage, ImageAspectRatio?)
```
```swift
enum ImageAspectRatio {
        case height(CGFloat, ImageAspectRatioType)
        case width(CGFloat, ImageAspectRatioType)
}

enum ImageAspectRatioType {
        case oneOne
        case sixteenNine
        case sevenTen
}
```

Maybe it should be decided by Design team

DavidMarinCalleja:
> **Screenshot tests report**
> ❌ ControlsTests / testTabBarControl()

It looks like a flaky test

Sources/Mistica/Components/Lists/ListCellContentView.swift
DavidMarinCalleja: Maybe we could start with static types and if in the future it is necessary we can add a `custom`case. 
```
enum ImageAspectRatioType {
     case oneOne
     case sixteenNine
     case sevenTen
     case custom(width: CGFloat, height: CGFloat)
}
```

Sources/Mistica/Components/Lists/Internals/CellLeftSectionView.swift
```diff
@@ -84,40 +84,48 @@ private extension CellLeftSectionView {
...
 
 private extension ListCellContentView.CellAssetType {
-    var assetSize: CGFloat {
+    var assetSize: CGSize {
         switch self {
         case .none:
-            return 0
-        case .image:
-            return ImageSize.large
+            return CGSize(width: 0, height: 0)
```
DavidMarinCalleja, line 90:
```suggestion
            return CGSize.zero
```

Sources/Mistica/Components/Lists/Internals/CellLeftSectionView.swift
```diff
@@ -84,40 +84,48 @@ private extension CellLeftSectionView {
...
     }
 
-    var viewSize: CGFloat {
+    var viewSize: CGSize {
         switch self {
         case .none:
-            return 0
-        case .image, .largeIcon:
-            return ImageSize.large
+            return CGSize(width: 0, height: 0)
```
DavidMarinCalleja, line 102:
```suggestion
            return CGSize.zero
```

Sources/Mistica/Components/Lists/Internals/CellLeftSectionView.swift
alejandroruizponce: Fixed!

Sources/Mistica/Components/Lists/Internals/CellLeftSectionView.swift
alejandroruizponce: Fixed!

alejandroruizponce:
> Thanks for the explanation! It looks great!! I miss:
> 
> 1. ~changes in SwiftUI Cell~
> 2. Add this configuration to our catalog to test it
>    Captura de pantalla 2023-11-13 a las 15 54 32
> 3. Add a couple of snapshot tests

Snapshot tests added and set in Catalog an image with default size to test it.

amegias (approved):
Good work Álex!

DavidMarinCalleja (approved):

PR #324, opened 2023-12-04
IOS-9604 Added infinite interval to crouton and snackbar for the Swif…

idenjoe:
## 🎟️ **Jira ticket**
[IOS-9604](https://jira.tid.es/browse/IOS-9604) Add infinite interval to Crouton/Snackbar SwiftUI versions.

## 🥅 **What's the goal?**
Add the same behaviour as in UIKit for crouton/snackbar SwiftUI versions with the infinite time invernal.

## 🚧 **How do we do it?**
1. Add a new time interval enum (fiveSeconds, tenSeconds, and infinite)
2. Added a block to return the dismiss reason(dismiss, button or timeout)
3. Modified the structure of Snackbar view to fit with the new close button cases.

Launch the Snackbar SwiftUI version in the catalog and show it with different settings.

## 🏑 **AppCenter build**

qr


Sources/MisticaSwiftUI/Components/Snackbar/View+Snackbar.swift
```diff
@@ -9,23 +9,44 @@
...
+        switch self {
+        case .fiveSeconds:
+            return 5
+        case .tenSeconds:
+            return 10
+        case .infinite:
+            return nil
+        }
+    }
+}
```
dhidalgofadrique, line 27: Maybe a more generic approach to ease adding more intervals in the future?

```suggestion
public enum CroutonDismissInterval {
    case seconds(TimeInterval)
    case infinite

    var timeInterval: TimeInterval? {
        switch self {
        case .seconds(let value):
            return value
        case .infinite:
            return nil
        }
    }
}
```

Sources/MisticaSwiftUI/Components/Snackbar/View+Snackbar.swift
```diff
@@ -166,3 +196,31 @@ private struct SnackbarModifier: ViewModifier {
...
+        case .none:
+            return .fiveSeconds
+        case .fiveSeconds:
+            return .fiveSeconds
+        case .tenSeconds:
+            return .tenSeconds
+        case .infinite:
+            return .infinite
+        }
+    }
```
dhidalgofadrique, line 218: Is this behavior specified in any document? It's confusing create a crouton with a dismiss interval of 5 seconds that dissappear after 10 seconds...

dhidalgofadrique:
In the Mistica Catalog the crouton is being shown under the tabbar 
Simulator Screenshot - iPhone 14 - 2023-12-05 at 08 55 01

Sources/MisticaSwiftUI/Components/Snackbar/Snackbar.swift
```diff
@@ -25,38 +25,105 @@ public enum SnackbarButtonStyle {
     case short
 }
 
-public struct Snackbar<Button: View>: View {
+public enum SnackbarDismissReason: Int, RawRepresentable {
```
amegias, line 28: could it be moved to MisticaCommon to avoid having the UIKit & SwiftUI representations?

Sources/MisticaSwiftUI/Components/Snackbar/Snackbar.swift
```diff
@@ -25,38 +25,105 @@ public enum SnackbarButtonStyle {
...
+    case dismiss
+    case button
+    case timeout
+
+    public typealias RawValue = String
```
amegias, line 33: weird...
what about having?
```diff
- public enum SnackbarDismissReason: Int
+ public enum SnackbarDismissReason: String
```

Then, you can assign its `String` directly
```swift
    case dismiss = "DISMISS"
    case button = "BUTTON"
    case timeout = "TIMEOUT"
```
you will be able to remove `var rawValue` and `init?(rawValue: RawValue)`

Sources/MisticaSwiftUI/Components/Snackbar/View+Snackbar.swift
```diff
@@ -9,23 +9,44 @@
 import Foundation
 import SwiftUI
 
+public enum CroutonDismissInterval {
```
amegias, line 12: Repeated in UIKit, could we put it in common?

Sources/MisticaSwiftUI/Components/Snackbar/View+Snackbar.swift
amegias: our components have a Readme so we can include it to explain this. Good idea!

Sources/MisticaSwiftUI/Components/Snackbar/View+Snackbar.swift
```diff
@@ -166,3 +196,31 @@ private struct SnackbarModifier: ViewModifier {
...
+            return .infinite
+        }
+    }
+
+    func dismiss(with reason: SnackbarDismissReason) {
+        DispatchQueue.main.async {
+            withAnimation { isVisible = false }
+            dismissHandlerBlock?(reason)
+        }
+    }
```
amegias, line 225: I THINK you can do:
```suggestion
    @MainActor
    func dismiss(with reason: SnackbarDismissReason) {
            withAnimation { isVisible = false }
            dismissHandlerBlock?(reason)
    }
```

Tests/MisticaSwiftUITests/UI/CroutonTests.swift
```diff
@@ -109,4 +109,55 @@ final class CroutonTests: XCTestCase {
...
+        let view = Rectangle()
+            .foregroundColor(.white)
+            .crouton(
+                isVisible: .constant(true),
+                title: "Title",
+                buttonTitle: "Action",
+                buttonAction: {},
+                forceDismiss: true
+            )
+        assertSnapshot(
```
amegias, line 123: thanks for adding tests and avoiding generating several brands for this :)!!

MisticaCatalog/Source/Catalog/MisticaSwiftUI/Components/SnackbarCatalogView.swift
```diff
@@ -18,7 +18,10 @@ struct SnackbarCatalogView: View {
     @State var buttonStyles: [SnackbarButtonStyle] = [.short, .large]
     @State var presentingSnackbar = false
     @State var presentingCrouton = false
-    @State var autoDismissDelay = 3
+    @State var autoDismissDelay: CroutonDismissInterval = .fiveSeconds
+    @State var intervalStyles: [CroutonDismissInterval] = [.fiveSeconds, .tenSeconds, .infinite]
```
amegias, line 22: it's a constant so it shouldn't be `@State`.
You can simplify this removing it and:
```diff
- autoDismissDelay: intervalStyles[selectedIntervalStyleIndex],
+ autoDismissDelay: CroutonDismissInterval.allCases[selectedIntervalStyleIndex],
```
wdyt?

Sources/MisticaSwiftUI/Components/Snackbar/View+Snackbar.swift
idenjoe: We can't do this because you let the user to add any interval and its definition does not allow this.

amegias:
Is it working properly?

Captura de pantalla 2023-12-05 a las 11 08 22

idenjoe:
> Is it working properly?
> 
> Captura de pantalla 2023-12-05 a las 11 08 22

Yes. The close button only should be added when the the time interval is infinite AND the action is nil OR the action is not nil and the forceDismiss is enabled.

Sources/MisticaSwiftUI/Components/Snackbar/Snackbar.swift
idenjoe: I moved the reason to MisticaCommon and now we need to mark the enum as @objc (it's being used from some methods marked as @objc) so we can not use String as their type.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 
 ### BREAKING CHANGES
 
-* **crouton:** DismissHandlerBlock has changed and now has a CroutonControllerDismissReason as return parameter
+* **crouton:** DismissHandlerBlock has changed and now has a SnackbarDismissReason as return parameter
```
amegias, line 11: don't change this, it's generated automatically ;)

amegias:
Should we extend this https://github.com/Telefonica/mistica-ios/blob/main/Sources/Mistica/Components/Crouton/README.md ?

Maybe something like:
### Config
You can use several intervals with different features:
| interval | Close | Action | ... |
|-------------|-------------|-------------|------------|
| 5"     | NO     | NO | ... |
| 10"     | Optional     | Mandatory | ... |
| ...     | ...     | ... | ... |

Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
```diff
@@ -68,20 +36,16 @@ public extension CroutonController {
     @available(iOSApplicationExtension, unavailable)
     @discardableResult
     func showCrouton(
-        withText text: String,
-        action: ActionConfig? = nil,
+        config: SnackbarConfig,
         style: CroutonStyle = .info,
```
dhidalgofadrique, line 40: Nice! Managing one element with all the info, instead of multiple parameters

MisticaCatalog/Source/Catalog/MisticaSwiftUI/Components/SnackbarCatalogView.swift
```diff
@@ -105,14 +131,16 @@ struct SnackbarCatalogView: View {
...
 extension SnackbarCatalogView {
-    func timeIntervalDescription(from interval: CroutonDismissInterval) -> String {
+    func timeIntervalDescription(from interval: SnackbarDismissInterval) -> String {
         switch interval {
         case .fiveSeconds:
             return "5"
         case .tenSeconds:
             return "10"
-        case .infinite:
+        case .infinity(_):
```
dhidalgofadrique: If parameter is not needed, then it's enough with:

`case .infinity:`

dhidalgofadrique (approved):
👍 SwiftUI Skills +1 😂

PR #328, opened 2024-01-15
fix(TabsView): Allows consumers to select an item and deselect all

crisrojas:
## 🎟️ **Jira ticket**

https://jira.tid.es/browse/SMARTWIFI-8301

## 🥅 **What's the goal?**

Allow users of TabsView component to select a tabItem when needed.

This is needed in SmartWiFi to allow tabItem selection on controller swipe when used in complement with a UIPageController.

See following video:

https://github.com/Telefonica/mistica-ios/assets/23734068/5bb22d7f-6676-4fb8-802b-895511330a7d

## 🚧 **How do we do it?**
- Add `deselectAll` method as we need to previously deselect all items before selecting the target controller on swipe (this prevents cumulating indicators in tabs)
- Make `selectTabItem(at:)` public so we can use it from outside.

See previously attached video.

### DIFF
--- a/Sources/Mistica/Components/Tabs/TabsView.swift
+++ b/Sources/Mistica/Components/Tabs/TabsView.swift
@@ -105,7 +105,7 @@ public extension TabsView {
             self.selectTabItem(at: max(0, index - 1))
         }
     }
-    
+
     func selectTabItem(at row: Int) {
         let indexPath = IndexPath(item: row, section: 0)
         firstIndexPathForSelectedItem = indexPath
@@ -119,9 +119,9 @@ public extension TabsView {
         let tabItem = tabsItems[row]
         delegate?.tabsView(self, didSelectTab: tabItem)
     }
-    
+
     func deselectAll() {
-        for indexPath in (collectionView.indexPathsForSelectedItems ?? []) {
+        for indexPath in collectionView.indexPathsForSelectedItems ?? [] {
             deselectTabItem(at: indexPath.row)
         }
     }
--- a/Sources/Mistica/Components/Tabs/TabsView.swift
+++ b/Sources/Mistica/Components/Tabs/TabsView.swift
@@ -107,6 +107,7 @@ public extension TabsView {
     }
 
     func selectTabItem(at row: Int) {
+        deselect()
         let indexPath = IndexPath(item: row, section: 0)
         firstIndexPathForSelectedItem = indexPath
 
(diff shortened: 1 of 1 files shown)


Sources/Mistica/Components/Tabs/TabsView.swift
```diff
@@ -105,6 +105,26 @@ public extension TabsView {
...
+        collectionView.selectItem(at: indexPath, animated: true, scrollPosition: .centeredHorizontally)
+
+        let tabItem = tabsItems[row]
+        delegate?.tabsView(self, didSelectTab: tabItem)
+    }
+
+    func deselectAll() {
+        for indexPath in collectionView.indexPathsForSelectedItems ?? [] {
+            deselectTabItem(at: indexPath.row)
+        }
```
amegias, line 126: I would call `deselect` since `All` looks like we can select several tabs and it's not possible, right?

crisrojas: Nice catch 👍

Sources/Mistica/Components/Tabs/TabsView.swift
```diff
@@ -105,6 +105,26 @@ public extension TabsView {
...
+
+        if let tabItemView = collectionView.cellForItem(at: indexPath) as? TabItemViewCell {
+            tabItemView.showSelected()
+        }
+
+        collectionView.selectItem(at: indexPath, animated: true, scrollPosition: .centeredHorizontally)
+
+        let tabItem = tabsItems[row]
+        delegate?.tabsView(self, didSelectTab: tabItem)
+    }
```
amegias, line 121: what happens if the dev calls several times to this method?
Example:
```swift
selectTabItem(at: 0)
selectTabItem(at: 1)
```
We should select **only** the tab at **index 1**, right? could you check it? if not, we should protect it

crisrojas: We would accummulate indicators, I thought about calling the `deselect` method inside the body of this function, but I ended up doing it from the outside. Should I go that way ?

amegias: Checking the spec
https://www.figma.com/file/tOb2OmRf6gnEK3fnSWPKOG/%F0%9F%94%B8-Tabs-Specs?type=design&node-id=0-1&mode=design&t=QKMt69gffQaa5UOn-0
I wouldn't allow selecting several tabs...

crisrojas: done 👍

amegias (approved):

WanaldinoTelefonica (approved):

PR #330, opened 2024-01-17
IOS-9612 Add Record Screenshots GH action

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9612

## 🥅 **What's the goal?**
The goal is to allow through a GH action to record new screenshots of tests that need it because they present a difference with the current ones and give error. These new screenshots are generated and auto-committed in the branch where the action is launched.

## 🚧 **How do we do it?**
To be able to generate the new screenshots we would need:
- Enable screenshots recording in each test through `isRecording = true`
- Launch the make test command that will run all the tests and those that fail as they will have the `isRecording` flag enabled new screenshots will be generated only for them. 
- Disable the screenshots recording.
- Commit new screenshots in the branch.

In the second step, the make test step, it is necessary to force the workflow to continue ignoring any errors in this step through the flag: `continue-on-error: true`

Apart from this workflow, you will see that in the PR there are many new screenshots that we are going to integrate. These screenshots have been generated from the two slaves (3 and 4) probably because the old ones were generated for a simulator, an iOS or from a different processor than the one defined in the Makefile of the slaves.

Once these screenshots are integrated, the future screenshots generated by the action should be exclusively those that fail in the development of our branch.

**_Once the PR is approved, I will re-establish the workflow we used for testing and move everything to a new workflow file._**

Right now it is for testing from the workflow action called "Generate Mistica Tokens" from the branch of this PR.
 **Important to use the branch of this PR, any other branch will launch the normal Mistica Tokens workflow instead of ours.**


.github/workflows/generate-mistica-tokens.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Commit Changes
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          commit_message: Regenerate screenshots automatically from GH
+          commit_message: Record screenshots automatically launched from GH action
```
amegias, line 33: I would add:
`file_pattern: '*.png'` => to avoid adding something else

.github/workflows/generate-mistica-tokens.yml
```diff
@@ -1,11 +1,11 @@
...
+name: Generate Mistica Tokens
 
 on:
   workflow_dispatch:
 
 jobs:
-  regenerate-screenshots:
-    name: Regenerate screenshots
+  record-screenshots:
+    name: Record screenshots
```
amegias, line 8: not sure if we should protect main to avoid adding changes there... 🤔 🤔 🤔 🤔

amegias:
TIP: Next time, add a single commit with the new screenshots. Then the reviewers will be able to select all commits except that one and their browser will be able to load the PR 😅 😅 😅 
Captura de pantalla 2024-01-18 a las 9 52 55

My browser trying to load 664 files:

.github/workflows/generate-mistica-tokens.yml
alejandroruizponce: Good idea, added! 👍

.github/workflows/generate-mistica-tokens.yml
alejandroruizponce: Mm, maybe I could add a check branch before the auto-commit step, if the branch is main the workflow is terminated and if not, it continues.

.github/workflows/generate-mistica-tokens.yml
```diff
@@ -2,30 +2,37 @@ name: Generate Mistica Tokens
...
       - name: Checkout
         uses: actions/checkout@v3             
 
-      - name: Generate Skin
-        run: make skin ref=${{github.event.inputs.ref}}
+      - name: Enable screenshots recording
+        run: |
+          find Tests/MisticaCommonTests -name "*.swift" -exec sed -i '' 's/isRecording = false/isRecording = true/' {} +
+          find Tests/MisticaSwiftUITests -name "*.swift" -exec sed -i '' 's/isRecording = false/isRecording = true/' {} +
+          find Tests/MisticaTests -name "*.swift" -exec sed -i '' 's/isRecording = false/isRecording = true/' {} +
```
idenjoe, line 18: Is there the possibility of enabling recording only for those that fail?

.github/workflows/generate-mistica-tokens.yml
amegias: I would add it **at the beginning** in order to avoid executing steps to do nothing ;)

.github/workflows/generate-mistica-tokens.yml
alejandroruizponce: I have been looking at it but it complicates the workflow a lot, you have to look for the files by going through the test functions that have failed.... It's confusing in my opinion.  What I have done is to leave everything in a single line so that it is valid in any project.

amegias (approved):

alexanegon (approved):

idenjoe (approved):

PR #332, opened 2024-01-25
IOS-9726 Fix close button with any duration

idenjoe:
## 🎟️ **Jira ticket**
[IOS-9726](https://jira.tid.es/browse/IOS-9726) Fix a bug where the close button can only be configured with infinite duration

## 🥅 **What's the goal?**
The crouton/scnakbar should be able to show the close button with any duration configured.

## 🚧 **How do we do it?**
Remove the infiniteWithClose interval option and pass a flag which defines if the close button should be forced or not( if the duration is infinite and has no action defined, the close button will also be presented).

You can test it in the catalog.

### DIFF
--- a/Sources/Mistica/Components/Crouton/Presentation/CroutonView.swift
+++ b/Sources/Mistica/Components/Crouton/Presentation/CroutonView.swift
@@ -24,7 +24,7 @@ class CroutonView: UIView {
         static let margins = NSDirectionalEdgeInsets(top: 14, leading: 16, bottom: 14, trailing: 16)
         static let marginsWhenUsingSafeArea = NSDirectionalEdgeInsets(top: 14, leading: 16, bottom: 0, trailing: 16)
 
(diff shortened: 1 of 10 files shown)


Sources/Mistica/Components/Crouton/Presentation/CroutonView.swift
```diff
@@ -24,7 +24,7 @@ class CroutonView: UIView {
         static let margins = NSDirectionalEdgeInsets(top: 14, leading: 16, bottom: 14, trailing: 16)
         static let marginsWhenUsingSafeArea = NSDirectionalEdgeInsets(top: 14, leading: 16, bottom: 0, trailing: 16)
 
-        static let buttonWidthThresholdForVerticalLayout: CGFloat = 104
+        static let buttonWidthThresholdForVerticalLayout: CGFloat = 128
```
idenjoe, line 27: This is another fix related with the size from where we need to put the action button above the message.

PR #337, opened 2024-02-20
feat(Callout): IOS-9818 Adding inverse variant

salavert:
## 🎟️ **Jira ticket**
[IOS-9818](https://jira.tid.es/browse/IOS-9818) Callout inverse variant

## 🥅 **What's the goal?**
We need to behave like Android and allow an inverse (bakground) variant that will be enabled by default with a **backgroundContainer** color, and **backgroundAlternative** when not enabled.

## 🚧 **How do we do it?**
See Android [commit](https://github.com/Telefonica/mistica-android/commit/069dbd6b5ebf0d08d4e0137250ded440791e7f13#diff-c40f0e2acf9b132af8b3f480beeb268b726b561b932380e950e883b0326f9babR163)

### DIFF
--- a/Sources/Mistica/Components/Callout/CalloutConfiguration.swift
+++ b/Sources/Mistica/Components/Callout/CalloutConfiguration.swift
@@ -49,7 +49,7 @@ public struct CalloutConfiguration {
         description: String,
         actions: CalloutActions? = nil,
         canClose: Bool = false,
-        inverse: Bool = true
+        inverse: Bool = false
     ) {
         self.asset = asset
         self.title = title
--- a/Sources/Mistica/Components/Callout/CalloutConfiguration.swift
+++ b/Sources/Mistica/Components/Callout/CalloutConfiguration.swift
@@ -33,6 +33,7 @@ public struct CalloutConfiguration {
     let description: String
     let actions: CalloutActions?
     let canClose: Bool
+    let inverse: Bool
 
     /// Callout Configuration to setup the component.
     /// - Parameters:
@@ -41,11 +42,20 @@ public struct CalloutConfiguration {
     ///   - description: This value is mandatory and will be printed as the description title of the callout
     ///   - actions: To know which buttons we must have inside the callout
     ///   - canClose: If true we will have a close button in the top-right to close the callout with an animation
-    public init(asset: CalloutAssetType = .none, title: String? = nil, description: String, actions: CalloutActions? = nil, canClose: Bool = false) {
+    ///   - inverse: If true we will have an inverse variant for background color
+    public init(
+        asset: CalloutAssetType = .none,
+        title: String? = nil,
+        description: String,
+        actions: CalloutActions? = nil,
+        canClose: Bool = false,
+        inverse: Bool = true
+    ) {
         self.asset = asset
         self.title = title
         self.description = description
         self.actions = actions
         self.canClose = canClose
+        self.inverse = inverse
     }
 }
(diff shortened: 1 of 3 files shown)


alexanegon (approved):

Sources/Mistica/Components/Callout/CalloutConfiguration.swift
```diff
@@ -41,11 +42,20 @@ public struct CalloutConfiguration {
...
     ///   - canClose: If true we will have a close button in the top-right to close the callout with an animation
-    public init(asset: CalloutAssetType = .none, title: String? = nil, description: String, actions: CalloutActions? = nil, canClose: Bool = false) {
+    ///   - inverse: If true we will have an inverse variant for background color
+    public init(
+        asset: CalloutAssetType = .none,
+        title: String? = nil,
+        description: String,
+        actions: CalloutActions? = nil,
+        canClose: Bool = false,
+        inverse: Bool = true
```
haynlo, line 52: Personal taste. If you want to make this param optional (in [Android](https://github.com/Telefonica/mistica-android/blob/main/library/src/main/java/com/telefonica/mistica/compose/callout/Callout.kt#L40) it is not), it should behave with default skin (not inverse), so maybe it would be better to set `inverse: Bool = false` instead.

haynlo (approved):
Nice!

Sources/MisticaSwiftUI/Components/Callout/Callout.swift
```diff
@@ -101,10 +104,14 @@ public struct Callout<LeadingButton: View, TrailingButton: View>: View {
             }
         }
         .padding(16)
-        .background(Color.backgroundAlternative)
+        .background(backgroundColor)
         .round(radiusStyle: .container)
     }
 
+    private var backgroundColor: Color {
+        inverse ? .backgroundContainer : .backgroundAlternative
```
alexanegon, line 112: if inverse var is true, it must return the alternative color, mustn't it?

salavert: This is simply behaving like Android
Screenshot 2024-02-20 at 16 19 11

salavert: @haynlo maybe this should be reversed?

Sources/Mistica/Components/Callout/CalloutConfiguration.swift
salavert: Good idea, done

Sources/MisticaSwiftUI/Components/Callout/Callout.swift
haynlo: According to [Figma spec](https://teams.microsoft.com/l/message/19:4cd911e0677a499b8c31c1311a273a3b@thread.skype/1708090237606?tenantId=9744600e-3e04-492e-baa1-25ec245c6f10&groupId=bef9be07-03e4-4afb-8a79-f720a63e4c28&parentMessageId=1707842587943&teamName=Novum&channelName=App%20Builder%20(formerly%20CPS)-public&createdTime=1708090237606), Inverse is not related to Alternative values. For Telefonica brand TOKENS, [backgroundAlternative](https://github.com/Telefonica/mistica-design/blob/production/tokens/telefonica.json#L48) is Grey but [backgroundContainer](https://github.com/Telefonica/mistica-design/blob/production/tokens/telefonica.json#L13) is White, that's why it is defined in that way in Android. 

Inverse callout should always have to be with White background according to some extra info I've found in [Teams thread](https://teams.microsoft.com/l/message/19:4cd911e0677a499b8c31c1311a273a3b@thread.skype/1708090237606?tenantId=9744600e-3e04-492e-baa1-25ec245c6f10&groupId=bef9be07-03e4-4afb-8a79-f720a63e4c28&parentMessageId=1707842587943&teamName=Novum&channelName=App%20Builder%20(formerly%20CPS)-public&createdTime=1708090237606).

salavert: Thanks for the explanation, it's clear for me

alexanegon (approved):

PR #341, opened 2024-03-04
IOS-9656 Add SVG support in UIButtons images

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9656

## 🥅 **What's the goal?**
To give support to be able to download images from URL that are in SVG format and that we can configure them as image in components like UIButton. This will allow that in apps like Novum we can configure the navigation bar buttons with SVG URLs.

## 🚧 **How do we do it?**
For this purpose, a setImageFromURL method has been implemented in a public extension of UIButton. This method will make use of the SDWebImage library that allows us to download images in SVG format and set them inside the UIButton image.

### DIFF
--- a/Sources/Mistica/Utils/Extensions/UIBarButtonItem+Utils.swift
+++ b/Sources/Mistica/Utils/Extensions/UIBarButtonItem+Utils.swift
@@ -51,30 +51,25 @@ extension UIButton {
 }
 
 public extension UIButton {
+    /* 
+     Support SVG Images from URL for UIButton settings
+     */
     func setImageFromURL(url: URL, urlForDarkMode: URL? = nil) {
-        print("TEST - setImageFromURL: \(url)")
         let coder = SDImageSVGCoder.shared
         SDImageCodersManager.shared.addCoder(coder)
 
-        let imageAsset = UIImageAsset()
-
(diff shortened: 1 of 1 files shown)


Sources/Mistica/Utils/Extensions/UIBarButtonItem+Utils.swift
```diff
@@ -47,3 +49,28 @@ extension UIButton {
         frameHeight = 36.0
     }
 }
+
+public extension UIButton {
+    /*
+     Support SVG Images from URL for UIButton settings
+     */
+    func setImageFromURL(url: URL, urlForDarkMode: URL? = nil) {
```
WanaldinoTelefonica, line 57:
```suggestion
    func setImage(from url: URL, darkURL: URL? = nil) {
```
I think this follows better the guidelines of swift, wdyt?

WanaldinoTelefonica (approved):

franciscojrp (approved):

Sources/Mistica/Utils/Extensions/UIBarButtonItem+Utils.swift
alejandroruizponce: Since it is a method expressly made for iphoneApp, there we are using urlDarkMode so it is better to keep it like this for similarity.

PR #348, opened 2024-03-26
Add default ids for listCellContentView and fix callout component ids

cbarcom:
## 🎟️ **Jira ticket**
[IOS-9930](https://jira.tid.es/browse/IOS-9930)

## 🥅 **What's the goal?**
List rows don't have default accessibility identifiers, so they have been assigned.
Also, according to Figma callout ids, they were incorrect and this PR fix them by removing section value from their id. Now, 
DefaultAccessibilityIdentifier.Section will be optional.

### DIFF
--- a/Sources/Mistica/Components/Callout/Callout.swift
+++ b/Sources/Mistica/Components/Callout/Callout.swift
@@ -131,7 +131,7 @@ public extension Callout {
             closeImageView.accessibilityIdentifier
         }
         set {
-            closeImageView.accessibilityIdentifier = newValue == nil ? CalloutAccessibilityIdentifiers.closeButton.rawValue : newValue
+            closeImageView.accessibilityIdentifier = newValue ?? CalloutAccessibilityIdentifiers.closeButton.rawValue
         }
     }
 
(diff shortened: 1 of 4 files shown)


Gomay88 (approved):

Sources/Mistica/Components/Callout/Callout.swift
```diff
@@ -126,6 +126,15 @@ public extension Callout {
         }
     }
 
+    var closeIdentifier: String? {
+        get {
+            closeImageView.accessibilityIdentifier
+        }
+        set {
+            closeImageView.accessibilityIdentifier = newValue == nil ? CalloutAccessibilityIdentifiers.closeButton.rawValue : newValue
```
DavidMarinCalleja, line 134:
```suggestion
            closeImageView.accessibilityIdentifier = newValue ?? CalloutAccessibilityIdentifiers.closeButton.rawValue
```

DavidMarinCalleja (changes requested):

DavidMarinCalleja (approved):

PR #350, opened 2024-04-01
IOS-9941 Update SDKs with Privacy Manifest

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9941

## 🥅 **What's the goal?**
We are updating all the third party SDKs we have because we need them to include the Privacy Manifest that Apple is now requiring for greater transparency in the privacy of both our app and the SDKs we use.

In this PR it is time to update Lottie and SDWebImage SDKs.

For more information:
https://confluence.tid.es/pages/viewpage.action?spaceKey=CTO&title=%5BiOS%5D+Privacy+manifest

## 🚧 **How do we do it?**
When upgrading Lottie it has been necessary to make some changes:
- `AnimationView` variable has been renamed to `LottieAnimationView`.
- `Animation` variable has been renamed to `LottieAnimation`.

For the update of SDWebImage no changes have been necessary.

PR #353, opened 2024-04-03
IOS-9965 Fix bounds in StepView animation

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-9965

## 🥅 **What's the goal?**
It has been detected that in the previous PR(https://github.com/Telefonica/mistica-ios/pull/350) on the Lottie update a collateral bug occurred that had not been resolved in the StepView, in its animation. This animation appeared to be minimally but erroneously cut off at the sides.

Captura de pantalla 2024-04-03 a las 16 10 29

## 🚧 **How do we do it?**
To fix it and leave it as it was before the update, the line has been added:
`animation.maskAnimationToBounds = false`

So that it is not clipped in its bounds. The snapshots tests will be re-generated in this PR to correct them.

It can be verified with the new snapshots and also a version of the Catalog has been generated in case you want to verify that the StepView is displayed correctly.


dhidalgofadrique (approved):
Well done! 👍

franciscojrp (approved):

PR #357, opened 2024-04-11
IOS-10003 Fix color token importers with type filter

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-10003

## 🥅 **What's the goal?**
Due to the future new gradient tokens and as an improvement to the robustness of the color token importers it is necessary to download only those tokens from the color palette that are of color type. If in the future the color tokens were of gradient type it would create an issue.

## 🚧 **How do we do it?**
Type checkers have been added to each template involved in the generation of color tokens.

It has been verified that the script is correct by running in terminal:
`make skin ref=production`

The result was that **the new generated color palette doesn't contain the existing gradient token** in the mistica-design json:
https://github.com/Telefonica/mistica-design/blob/e58edd73c4e4877fc8b902f8a2d9a31da5ca8b8a/tokens/movistar.json#L623

**The other tokens remain the same and as before.**

### DIFF
--- a/_templates/ColorTokenGenerator/BrandColors/BrandColorsAction.ejs.t
+++ b/_templates/ColorTokenGenerator/BrandColors/BrandColorsAction.ejs.t
@@ -37,8 +37,10 @@ struct <%= className %>Colors: MisticaColors {
 }
 
 public struct <%= className %>ColorPalette {
-	public init() {}
-	<%_ Object.keys(jsonObject.jsonData.global.palette).forEach(function(key) { -%>
-	public let <%= key %> = UIColor(hex:"<%= jsonObject.jsonData.global.palette[key].value %>")!
-	<%_ }); -%>
+    public init() {}
+    <%_ Object.keys(jsonObject.jsonData.global.palette).forEach(function(key) { -%>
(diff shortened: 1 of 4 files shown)


idenjoe (approved):

_templates/ColorTokenGenerator/BrandColors/BrandColorsAction.ejs.t
```diff
@@ -22,23 +22,27 @@ struct <%= className %>Colors: MisticaColors {
...
 }
 
 public struct <%= className %>ColorPalette {
-	public init() {}
-	<%_ Object.keys(jsonObject.jsonData.global.palette).forEach(function(key) { -%>
-	public let <%= key %> = UIColor(hex:"<%= jsonObject.jsonData.global.palette[key].value %>")!
-	<%_ }); -%>
+    public init() {}
+    <%_ Object.keys(jsonObject.jsonData.global.palette).forEach(function(key) { -%>
+        <%_ if (jsonObject.jsonData.global.palette[key].type === "color") { -%>
```
amegias, line 44: thanks!

amegias (approved):

PR #358, opened 2024-04-11
IOS-10010 Add new brand O2New

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-10010

## 🥅 **What's the goal?**
Add the new brand called **O2New** with all its new colors, fonts, radius, etc.

More info here: https://github.com/Telefonica/mistica-design/blob/production/tokens/o2-new.json

## 🚧 **How do we do it?**
For this purpose, the following steps have been taken:
- Firstly to be able to carry out this development without the token importer failing due to the fact that currently there are gradients inside the color tokens that are not of color type this PR has been carried out to solve this problem: 
https://github.com/Telefonica/mistica-ios/pull/357
- Makefile file has been updated to capture the values of the json file in O2New's mistica-design when we do a `make skin`.
- Once the Makefile has been updated, using the command `make skin ref=production` we have downloaded and created all the palettes related to colors, radius, fonts and others.
- In MisticaConfig we have added the new brand O2New and in the project assets we have added the same assets that existed for O2.

Also the palettes of the rest of the brands have been updated, it would not have been necessary if the MisticaColors protocol had not changed with new tokens, but unfortunately it has changed and it is not possible that O2New has those new tokens and the rest of the brands do not have them. For that reason we have taken the opportunity to update the rest of the palettes as well.

**More info here:** https://confluence.tid.es/display/CTO/%5BAPPS%5D+Shared+Spec%3A+Mistica+gradients

PR #361, opened 2024-04-22
IOS-10010 Add Gradients support in brands

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-10010

## 🥅 **What's the goal?**
Following the previous PR (https://github.com/Telefonica/mistica-ios/pull/358) where we introduced the new O2New brand, changes are required to support color tokens that instead of colors can be gradients indistinctly for the different brands. For that reason it's necessary to create a new supertype that can be color or gradient for a given token.

## 🚧 **How do we do it?**
In order to carry out this new implementation, the following important changes have been made:
- An enumerated has been created to act as a supertype called `MisticaColorStyle` that can be color or gradient. The gradient type will receive a `MisticaGradient` that has been defined according to what has been agreed with Design regarding the json that we will receive with the gradient data: https://github.com/Telefonica/mistica-design/blob/production/tokens/o2-new.json
- The two tokens affected at this moment by this supertype are only `backgroundBrand` and `backgroundContainerBrand`, and only in O2New its value is gradient type, in the rest of brands its value will remain color type.
- To be able to use the supertype in any view of our Mistica components it is necessary to create helpers that allow us both in UIKit and SwiftUI to receive a token type `MisticaColorStyle` and assign it to that view to decide whether to apply a color or a gradient to the background of the same.
- These helpers will use as base a `GradientView` **created exclusively in SwiftUI, the reason is that in UIKit the gradient generated differs from SwiftUI, its rendering is different, less progressive and therefore less similar to the one we want to achieve, which is a gradient similar to the web one.**
- Therefore for UIKit it is necessary to generate a UIViewController that contains a UIHostingController to encapsulate the SwiftUI `GradientView`. 
 - The Mistica components affected by this new supertype have been `Feedback`, `Header` and `List` cells. Several fixes had to be made so that the view generated by the UIHostingController would not give problems when configuring it in each view, which could be **a UIView or a UIStackView, or even a view cell. The solution must be valid for all views**.

From the Mistica Catalog **by choosing the O2New brand in the brand selector** it is possible to test the gradients on the following components:
- Feedback (.success type only)
- Header (Inverse style only)
- List (Inverse layout style only)

Demo test:


MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogFeedbacksViewController.swift
```diff
@@ -191,7 +191,8 @@ extension UICatalogFeedbacksViewController {
         switch indexPath.row {
         case 0:
             let navigationController = UINavigationController(rootViewController: feedbackViewController)
-            navigationController.navigationBar.applyMisticaStyle()
+            //navigationController.navigationBar.applyMisticaStyle()
```
alejandroruizponce: PENDING ISSUE HERE. Currently the Feedback component depends on a navigation bar whose color is separate from the background of the Feedback itself and should match so as not to break the UI of the Feedback. The problem is that it is not possible to define a gradient in two parts, one in the navigationBar and another in the Feedback. Design should decide what to do in this case.

MisticaCatalog/Source/Catalog/MisticaSwiftUI/Components/ButtonCatalogView.swift
```diff
@@ -67,7 +67,7 @@ struct ButtonsCatalogView: View {
 
                         Spacer()
                     }
-                    .listRowBackground(style.inverse ? Color.backgroundBrand : Color.background)
```
alejandroruizponce, line 70: This kind of backgrounds for Catalog or DEBUG tasks do not need to be backgroundBrand since they involve us using the helpers here and complicate the code for something that has no impact on production.

Sources/Mistica/Components/Header/HeaderView.swift
```diff
@@ -229,6 +229,7 @@ private extension HeaderView {
     }
 
     func updateTopStackView() {
+        bringSubviewToFront(stackView)
         topStackView.removeArrangedSubviews()
```
alejandroruizponce, line 233: When adding a subview in a UIStackView it is necessary to move to the front the arrenged subviews that we have so that they are not behind the gradient.

Sources/MisticaCommon/Components/GradientView+UIKit.swift
```diff
@@ -0,0 +1,94 @@
...
+                        .trailingAnchor
+                        .constraint(equalTo: view.trailingAnchor)
+            ])
+        
+
+    }
+}
+
+extension UIHostingController {
+    convenience public init(rootView: Content, ignoreSafeArea: Bool) {
```
alejandroruizponce: **WHY IS THIS NECESSARY?** Unfortunately I don't see any way to collide with the safe areas when adding GradientView in a full screen view such as Feedback. If we don't indicate to avoid the safe areas we are left with a blank space on the edges that I don't see any other way to avoid.

It cannot be put by default since for cases like the UIStackView of the Header or the own cells of List we would break the UI if we also ignore the safe areas there.

Sources/MisticaCommon/Components/GradientView.swift
```diff
@@ -0,0 +1,89 @@
...
+                gradient: Gradient(stops: stops),
+                startPoint: gradientPoints(size: geo.size, cssAngle: angle).start,
+                endPoint: gradientPoints(size: geo.size, cssAngle: angle).end
+            )
+        }
+    }
+}
+
+private extension GradientView {
+    func gradientPoints(size: CGSize, cssAngle: CGFloat) -> (start: UnitPoint, end: UnitPoint) {
```
alejandroruizponce, line 32: This function is intended to receive the same points that would be received on the Web using CSS from a given angle.

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogFeedbacksViewController.swift
```diff
@@ -191,7 +191,8 @@ extension UICatalogFeedbacksViewController {
         switch indexPath.row {
         case 0:
             let navigationController = UINavigationController(rootViewController: feedbackViewController)
-            navigationController.navigationBar.applyMisticaStyle()
+            // navigationController.navigationBar.applyMisticaStyle()
+            navigationController.navigationBar.isHidden = true
```
amegias, line 195: ????

amegias (changes requested):

Sources/MisticaCommon/Components/GradientView+UIKit.swift
```diff
@@ -0,0 +1,93 @@
...
+        hostingController.didMove(toParent: self)
+
+        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            hostingController.view
+                .centerXAnchor
+                .constraint(equalTo: view.centerXAnchor),
+            hostingController.view
+                .centerYAnchor
+                .constraint(equalTo: view.centerYAnchor),
```
amegias, line 46: are those two needed?

Sources/MisticaCommon/Components/GradientView.swift
amegias: I would add doc to this class explaining why we are using this view:
- to allow dark/light mode automatically

Sources/MisticaCommon/Components/GradientView.swift
```diff
@@ -0,0 +1,89 @@
...
+//  GradientView.swift
+//
+//  Made with ❤️ by Novum
+//
+//  Copyright © Telefonica. All rights reserved.
+//
+
+import SwiftUI
+
+public struct GradientView: View {
```
amegias, line 11: I would add doc to this class explaining why we are using this view:
- to allow dark/light mode automatically

```suggestion
/// ...
public struct GradientView: View {
```

Sources/MisticaCommon/Components/GradientView+UIKit.swift
```diff
@@ -0,0 +1,93 @@
...
+//  GradientView+UIKit.swift
+//
+//  Made with ❤️ by Novum
+//
+//  Copyright © Telefonica. All rights reserved.
+//
+
+import SwiftUI
+
+public class GradientSwiftUIViewController: UIViewController {
```
amegias, line 11: I would add a comment here to explain why you should use this hosting VC
```suggestion
/// ...
public class GradientSwiftUIViewController: UIViewController {
```

Sources/MisticaCommon/Colors/MisticaColorStyle.swift
```diff
@@ -0,0 +1,24 @@
...
+
+public extension MisticaColorStyle {
+    static var backgroundBrand: MisticaColorStyle {
+        MisticaConfig.currentColors.backgroundBrand
+    }
+
+    static var backgroundContainerBrand: MisticaColorStyle {
+        MisticaConfig.currentColors.backgroundContainerBrand
+    }
+}
```
amegias, line 24: needed? why?

Sources/MisticaCommon/Components/GradientView+UIKit.swift
amegias: I would explain it in the code to avoid forgetting this

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogFeedbacksViewController.swift
alejandroruizponce: Originally it was commented to see if the navigation bar was configured correctly or if it should be removed. Currently the color of the nav bar matches the background of the view so there is no difference that breaks the UI with the gradient. So these lines have been restaured.

Sources/Mistica/Utils/Extensions/UIView+Utils.swift
```diff
@@ -12,20 +12,20 @@ import UIKit
...
         case .gradient(let gradient):
-            applyLinearGradient(colors: gradient.colors, locations: gradient.stops, angle: gradient.angle, ignoreSafeArea: ignoreSafeArea)
+            addLinearGradient(colors: gradient.colors, locations: gradient.stops, angle: gradient.angle, ignoreSafeArea: ignoreSafeArea)
         }
     }
 
     // MARK: Linear gradient
 
-    public func applyLinearGradient(colors: [UIColor], locations: [CGFloat], angle: CGFloat, ignoreSafeArea: Bool) {
+    public func addLinearGradient(colors: [UIColor], locations: [CGFloat], angle: CGFloat, ignoreSafeArea: Bool) {
```
amegias, line 28: why don't you pass a `Gradient` var plus `ignoreSafeArea` var?

Sources/Mistica/Utils/Extensions/UIView+Utils.swift
```diff
@@ -6,11 +6,33 @@
...
+        case .color(let color):
+            backgroundColor = color
+        case .gradient(let gradient):
+            applyLinearGradient(colors: gradient.colors, locations: gradient.stops, angle: gradient.angle, ignoreSafeArea: ignoreSafeArea)
+        }
+    }
+
+    // MARK: Linear gradient
+
+    public func applyLinearGradient(colors: [UIColor], locations: [CGFloat], angle: CGFloat, ignoreSafeArea: Bool) {
```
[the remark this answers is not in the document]
alejandroruizponce: Yes, it's more accurated. Renamed.

Sources/MisticaCommon/Components/GradientView.swift
alejandroruizponce: Doc added in GradientSwifUIViewController, it's really which is gonna use this SwiftUI view.

Sources/MisticaCommon/Colors/MisticaColorStyle.swift
alejandroruizponce: As before we used extensions for UIColor and Color, this way we can directly instantiate backgroundBrand for a variable of type MisticaColor, like this one:
```
    var backgroundColor: MisticaColor {
        switch self {
        case .fullWidth:
            return .solid(.background)
        case .boxed:
            return .solid(.backgroundContainer)
        case .boxedInverse:
            return .backgroundBrand
        }
    }
```

Sources/MisticaCommon/Colors/MisticaColorStyle.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Copyright © Telefonica. All rights reserved.
+//
+
+import UIKit
+
+public enum MisticaColorStyle {
+    case color(UIColor)
+    case gradient(MisticaGradient)
+}
```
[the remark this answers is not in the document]
alejandroruizponce: I'm more convinced by the second option, renamed!

Sources/Mistica/Utils/Extensions/UIView+Utils.swift
alejandroruizponce: Well, I thought about it for the same reason that I have separated the addLinearGradient function from setMisticaColor, in case at some point we use the addLinearGradient func from another part of the app where we do not receive a Gradient object and the values are assigned manually. Maybe it won't happen. WDYT?

amegias: I have a no strong opinion here :/
I would take care of the current code, if you feel that passing a `Gradient` the current code will look better... go ahead

amegias (approved):

Sources/MisticaCommon/Colors/MisticaGradient.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Made with ❤️ by Novum
+//
+//  Copyright © Telefonica. All rights reserved.
+//
+
+import UIKit
+
+public struct MisticaGradient {
+    public let colors: [UIColor]
```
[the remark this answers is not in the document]
alejandroruizponce: Mm, but it's agreed with Design that this is the structure to follow. What would be the point if we had more colors than stops? Or vice versa? The gradient could not be applied correctly. I think this is a mandatory of the gradient structure, it should always be respected IMHO.

idenjoe (approved):

PR #364, opened 2024-04-30
Reverted change to fix /IOS-10056

dhidalgofadrique:
## 🎟️ **Jira ticket**
[IOS-10056](https://jira.tid.es/browse/IOS-10056) TopBarButtonItemPresenter.downloadIcons crash

## 🥅 **What's the goal?**
As part of [IOS-10056](https://jira.tid.es/browse/IOS-10056) fix which reverts changes done in https://github.com/Telefonica/mistica-ios/pull/350 we need to restore this code

## 🚧 **How do we do it?**
Restored `UIButton` extension method `setImageFromURL(url:urlForDarkMode:defaultImage:`

Just check the code

### DIFF
--- a/Sources/Mistica/Utils/Extensions/UIBarButtonItem+Utils.swift
+++ b/Sources/Mistica/Utils/Extensions/UIBarButtonItem+Utils.swift
@@ -49,3 +49,31 @@ extension UIButton {
         frameHeight = 36.0
     }
 }
+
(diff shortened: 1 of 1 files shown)


DavidMarinCalleja (approved):
thanks for fix it

idenjoe (approved):
Thx!

PR #375, opened 2024-06-25
IOS-10221 Accesibility of Snackbars and croutons on iOS for VIVO Novo App

dhidalgofadrique:
## 🎟️ **Jira ticket**
[IOS-10221](https://jira.tid.es/browse/IOS-10221) Accesibility of Snackbars and croutons on iOS for VIVO Novo App

## 🥅 **What's the goal?**
As part of [IOS-10221](https://jira.tid.es/browse/IOS-10221), it's necessary to make some changes in Mistica:

1. Be able to define the `accessibilityLabel` for crouton/snackbar action button
2. Made an accessibility announcement when a crouton/snackbar is presented

## 🚧 **How do we do it?**

1. Expose `accessibilityLabel` for crouton/snackbar actions, in order to be able to provide an accessibility text if needed to the action button
2. Added an `AccessibilityHelper` to encapsulate accessibility logic, in this particular case, to manage accessibility announecments.

Regarding to accessibility announecments, there are several limitations depending on iOS version:

### iOS16 and lower

In these versions of the OS, there is a mechanism to make accessibility announcements using `UIAccessibility.post(notification:argument:)` using `.announcement` as `notification` parameter and an attributed string with the attribute `. accessibilitySpeechQueueAnnouncement ` as argument. 

This attribute accepts a boolean values which indicates if the message should be enqueued if a current message is being announced (value of `true`) or if it has to be announced immediatly (value of `false`) interrupting a current announcement, but **_this priority mechanism is applied only between announcements made using `UIAccessibility.post`, if there is a current message being read by the system or a new one arrives, the custom announcement will be interrupted or won't be readed at all_**

### iOS17+

In iOS17 there is a new accessibility announcements mechanism using an `AttributedString` with the attribute `.accessibilitySpeechAnnouncementPriority`, having three possible values:

1. `low` -> Announcements are queued and spoken when other speech utterances have completed. If an OS speech arrives while speaking this, it will be interrupted by the OS one
2. `default` -> Announcements will interrupt existing speech, but are interruptible if a new speech utterance is started.
3. `high` -> Announcements will interrupt other speech (including OS ones) and cannot be interrupted once started.

After share this limiations with the accessibility chat, we decided the approach to use iOS17 functionality if available, using always high priority in order to ensure the announce is spoken. If not high, when presenting the crouton in PIN change view, it starts speaking it but when the view sets the focus to an element, the announcement is interrupted.

Here it is the discussion: https://teams.microsoft.com/l/message/19:cb3be251fe6e4ace8180c2e68f3112f4@thread.v2/1719229673024?context=%7B%22contextType%22%3A%22chat%22%7D

The best way to verify this is pointing iphoneapp to this branch and set/remove PIN code. Here are the examples in iOS16 and iOS17

| iOS16 | iOS17 |
|---|---|
| https://github.com/Telefonica/mistica-ios/assets/4386305/28a0b075-2dff-442c-aea7-056cf65c2dee | https://github.com/Telefonica/mistica-ios/assets/4386305/cc760c2a-4276-4653-8f87-086591029a7c |

## 🏑 **AppCenter build**
IOS-10221


idenjoe (approved):
Good job!

amegias (approved):

PR #386, opened 2024-07-17
feat(Snackbar): allow setting the accesibility identifier via SnackbarConfig

franciscojrp:
## 🎟️ **Jira ticket**
[WLLT-1516](https://jira.tid.es/browse/WLLT-1516)

## 🥅 **What's the goal?**
The current way of setting the accessibility identifier of the snarckbar title does not work for the SwiftUI modifier because the methods defined to set it are defined in Snackbar, but the SwiftUI modifier returns `some View` 

## 🚧 **How do we do it?**
Use the existing SnackbarConfig (used to set the title) to set also the accessibility label and identifier if needed

### DIFF
--- a/Sources/MisticaCommon/Components/Snackbar/Domain/SnackbarConfig.swift
+++ b/Sources/MisticaCommon/Components/Snackbar/Domain/SnackbarConfig.swift
@@ -72,11 +72,19 @@ public enum SnackbarDismissInterval: Equatable {
 
 public struct SnackbarConfig {
     public let title: String
+    public let titleAccessibilityLabel: String?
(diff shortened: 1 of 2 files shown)


pbartolome (approved):
I think we are not consistent with how we configure this view... some things are configurable with the config struct, and other things as parameters in the view modifier... should we change the view modifier instead? or move all configuration to the config struct? anyway this looks ok

franciscojrp:
This is probably used by a lot of projects, so I tried not to modify anything that would make all those projects to change its code

alejandroruizponce (approved):

PR #396, opened 2024-08-21
Update tokens

amegias:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-10491

## 🥅 **What's the goal?**
Taking advantage of https://github.com/Telefonica/mistica-ios/pull/397 which needs to release breaking changes, I've updated the tokens which, some of them, have been renamed (Selected -> Pressed) + added `title3` font size needed in that PR.

## 🚧 **How do we do it?**

 🪄 `make skin` 🪄 

- As you can see, tests still pass so nothing should be changed.
- The breaking changes are in this commit https://github.com/Telefonica/mistica-ios/pull/396/commits/f45624f077f1cd8474263a84782c05f4c32b0670
Since the color tokens are public, we will do something similar in each project (breaking changes)


yceballost (approved):
🧑‍🍳💋

salavert (approved):

alejandroruizponce (approved):
Great! Thanks

PR #402, opened 2024-10-03
IOS-10545 Update Makefile with XCode 16

alejandroruizponce:
## 🎟️ **Jira ticket**

## 🥅 **What's the goal?**
Build and integrate with XCode 16 and iOS 18 now we have slaves updated with that version.

## 🚧 **How do we do it?**
Update configuration in Makefile with a new simulator iPhone 16 in iOS 18.0 and using XCode 16 app.

Compiling with XCode 16 and on iPhone 16 misaligns some tiny pixels in the snapshots which requires that many of them have been re-generated with very little difference.

There are many tests that have been updated probably due to variations of 1px or less, this is something that already happened with iOS 17 back in the day, I recommend checking them from Sourcetree or similar as the Github page will freeze trying to load so many snapshots.

**For Design review**: A release for Mistica Catalog has been generated in Alpha Mistica-SwiftUI-iOS app center project.


Tests/MisticaTests/UI/__Snapshots__/ListsTests/testCellDefaultConfigAndLargeIconAndControlView.1.png
DavidMarinCalleja, line 1: it's weird but if you notice the new circle is painted worse than the old one. The old one has an “anti aliasing” effect and the new one has sawtooths in some areas.

Tests/MisticaTests/UI/__Snapshots__/ListsTests/testCellWithAllLabelsAndLargeIconAndControlView.1.png
DavidMarinCalleja, line 1: the same in this case, the circle is not pixel perfect

Tests/MisticaTests/UI/__Snapshots__/ListsTests/testCellDefaultConfigAndLargeIconAndControlView.1.png
alejandroruizponce: Yes, this is something that happens with XCode 16... I have seen in the snapshot testing repo that there are people with the same problem and it is not clear what the solution is... I have tried several things but I think I will have to wait for a future version of snapshot to fix this bug.

idenjoe (approved):

marinitx (approved):

PR #405, opened 2024-10-25
IOS-10545 Migrating to Swift 6.0 with strict concurrency

alejandroruizponce:
## 🎟️ **Jira ticket**

## 🥅 **What's the goal?**
In a previous PR we made changes to make Mistica work in XCode 16 without using Swift 6. Now in this PR Swift 6 is enabled and many new concurrency issues arise. You can check more information here:
https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/incrementaladoption

## 🚧 **How do we do it?**
- `Package` has been updated to support Swift 6.0 and the `SnapshotTesting` library has been configured to the latest version to support strict concurrency testing.
- Two elements closely related to all our Mistica logic such as **`UIView` and `View` are now executed in `@MainActor` which implies that all components related to these two classes must also be isolated in the main thread.**
- Classes like `FontStyle` or `MisticaConfig` that are quite important in Mistica's logic have been refactored as `@unchecked Sendable` and **concurrency will be handled manually through queues**. This way we do not isolate these classes which would mean that any place where they are used would also have to be isolated and that would be a very big refactor.
- All tests have been refactored with the changes in the latest version of SnapshotTesting, the most important change is:

Before in old version:
```
    override func setUp() {
        super.setUp()
        isRecording = false
    }
```

Now in latest version:
```
    override func invokeTest() {
        withSnapshotTesting(record: .never) {
            super.invokeTest()
        }
    }
```
_*This change implies changing our screenshots action since now the isRecording variable does not exist and must be configured in the new function. In fact this function now allows several values: **`.all, .failed, .missing, .never.`**_

**The screenshots record has been set to run with `.failed` in order to re-record only those screenshots that fail to be regenerated.**

A version of Catalog has been generated to verify that everything works as before.

### DIFF
--- a/.github/workflows/record-screenshots.yml
+++ b/.github/workflows/record-screenshots.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Enable screenshots recording
         run: |
-          find . -type f -name "*.swift" -exec sed -i '' 's/isRecording = false/isRecording = true/' {} +
+          find . -type f -name "*.swift" -exec sed -i '' 's/record: .never/record: .failed/' {} +
           
       - name: Launch tests and record screenshots
         run: make test
@@ -44,7 +44,7 @@ jobs:
         
       - name: Disable screenshots recording
         run: |
-          find . -type f -name "*.swift" -exec sed -i '' 's/isRecording = true/isRecording = false/' {} +
+          find . -type f -name "*.swift" -exec sed -i '' 's/record: .failed/record: .never/' {} +
 
       - name: Commit Changes
         uses: stefanzweifel/git-auto-commit-action@v5
--- a/Sources/Mistica/Components/Feedback/FeedbackView.swift
+++ b/Sources/Mistica/Components/Feedback/FeedbackView.swift
@@ -430,7 +430,7 @@ private extension FeedbackView {
                 Task { @MainActor in
                     self?.prepareAnimation()
                     self?.startAnimation()
-                    
+
                     self?.primaryButton?.title = title
                     self?.primaryButton?.isLoading = false
                 }
--- a/Sources/Mistica/Components/Feedback/FeedbackView.swift
+++ b/Sources/Mistica/Components/Feedback/FeedbackView.swift
@@ -215,7 +215,7 @@ public class FeedbackView: UIView {
         if let secondaryButton = secondaryButton {
             buttonsView.addArrangedSubview(secondaryButton)
         }
-        
+
         buttonsView.alignment = .fill
         buttonsView.axis = .vertical
         buttonsView.spacing = 16
--- a/Sources/Mistica/Components/Feedback/FeedbackView.swift
+++ b/Sources/Mistica/Components/Feedback/FeedbackView.swift
@@ -427,11 +427,13 @@ private extension FeedbackView {
             primaryButton?.isLoading = true
             prepareHapticFeedback()
             completion { [weak self] in
-                self?.prepareAnimation()
-                self?.startAnimation()
-
-                self?.primaryButton?.title = title
-                self?.primaryButton?.isLoading = false
+                Task { @MainActor in
+                    self?.prepareAnimation()
+                    self?.startAnimation()
+                    
+                    self?.primaryButton?.title = title
+                    self?.primaryButton?.isLoading = false
+                }
             }
         case .none:
             break
--- a/Sources/Mistica/Components/Feedback/FeedbackView.swift
+++ b/Sources/Mistica/Components/Feedback/FeedbackView.swift
@@ -208,11 +208,14 @@ public class FeedbackView: UIView {
     private lazy var buttonsView: UIView = {
         let buttonsView = UIStackView(arrangedSubviews: [])
 
-        Task { @MainActor in
-            try [primaryButton, secondaryButton].compactMap { $0 }
-                .forEach(buttonsView.addArrangedSubview(_:))
+        if let primaryButton = primaryButton {
+            buttonsView.addArrangedSubview(primaryButton)
         }
 
+        if let secondaryButton = secondaryButton {
+            buttonsView.addArrangedSubview(secondaryButton)
+        }
+        
         buttonsView.alignment = .fill
         buttonsView.axis = .vertical
         buttonsView.spacing = 16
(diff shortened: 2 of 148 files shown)


Sources/Mistica/Components/Feedback/FeedbackView.swift
```diff
@@ -208,8 +208,10 @@ public class FeedbackView: UIView {
     private lazy var buttonsView: UIView = {
         let buttonsView = UIStackView(arrangedSubviews: [])
 
-        [primaryButton, secondaryButton].compactMap { $0 }
-            .forEach(buttonsView.addArrangedSubview(_:))
+        Task { @MainActor in
+            try [primaryButton, secondaryButton].compactMap { $0 }
+                .forEach(buttonsView.addArrangedSubview(_:))
```
alejandroruizponce, line 213: Now addArrangedSubview is isolated:
`@MainActor open class UIStackView : UIView { .. }`

Sources/Mistica/Components/Feedback/FeedbackView.swift
```diff
@@ -365,7 +367,7 @@ private extension FeedbackView {
         }
 
         // Prepare
-        views.forEach(prepare(view:))
+        try? views.forEach(prepare(view:))
```
alejandroruizponce, line 370: Now `forEach` method has a rethrow to be handled:
`    @inlinable public func forEach(_ body: (Element) throws -> Void) rethrows
`

Sources/Mistica/Components/Feedback/FeedbackView.swift
```diff
@@ -410,8 +412,10 @@ private extension FeedbackView {
               let hapticFeedbackDelay = style.hapticFeedbackDelay else { return }
         Timer.scheduledTimer(withTimeInterval: hapticFeedbackDelay, repeats: false) { [weak self] _ in
             guard let self = self else { return }
-            self.feedbackGenerator?.notificationOccurred(hapticFeedbackStyle)
-            self.feedbackGenerator = nil
+            Task { @MainActor in
+                self.feedbackGenerator?.notificationOccurred(hapticFeedbackStyle)
+                self.feedbackGenerator = nil
+            }
```
alejandroruizponce, line 418: Now `UINotificationFeedbackGenerator` is isolated:
`@MainActor open class UINotificationFeedbackGenerator : UIFeedbackGenerator`

Sources/MisticaCommon/Extensions/Lottie+Utils.swift
```diff
@@ -15,7 +15,7 @@ import Lottie
  **/
 
 public extension Lottie.LottieConfiguration {
-    static var current: LottieConfiguration = {
+    nonisolated(unsafe) static var current: LottieConfiguration = {
```
alejandroruizponce, line 18: Lottie is not updated yet with strict concurrency.

.github/workflows/record-screenshots.yml
```diff
@@ -36,15 +36,15 @@ jobs:
 
       - name: Enable screenshots recording
         run: |
-          find . -type f -name "*.swift" -exec sed -i '' 's/isRecording = false/isRecording = true/' {} +
+          find . -type f -name "*.swift" -exec sed -i '' 's/record: .never/record: .failed/' {} +
```
salavert, line 39: What we do at Latch is 1 test plan with 3 configurations so we can specify them on `xcodebuild -testPlan LatchSnapshots -only-test-configuration "Record all"``

Screenshot 2024-10-28 at 11 28 50

salavert (approved):

.github/workflows/record-screenshots.yml
pbartolome: if you define this env variable, if only CI executes this and you don't want a test plan to do it locally, you should be able to provide it in the makefile and set it before executing the xcodebuild call

```
export RECORD_SCREENSHOTS=all
xcodebuild -project .... test
```

franciscojrp (approved):

Sources/Mistica/Components/Sheet/View/BottomSheetPresentation/BottomSheetInteractiveDismissalTransition.swift
```diff
@@ -11,7 +11,7 @@ import UIKit
 final class BottomSheetInteractiveDismissalTransition: NSObject {
     private enum Constants {
         static let maxBouncingHeight: CGFloat = 250
-        static let animationDuration: CGFloat = UIView.defaultAnimationDuration
+        static let animationDuration: CGFloat = 0.25
```
DavidMarinCalleja, line 14: Why this change?

Sources/MisticaSwiftUI/Utils/Extensions/EnvironmentValues.swift
```diff
@@ -21,7 +21,7 @@ public extension EnvironmentValues {
 @available(iOSApplicationExtension, unavailable)
 private struct SafeAreaInsetsKey: EnvironmentKey {
     static var defaultValue: EdgeInsets {
-        UIApplication.shared.keyWindow?.safeAreaInsets.swiftUiInsets ?? EdgeInsets()
+        EdgeInsets()
```
DavidMarinCalleja, line 24: Is this change right? I'm not sure we always want to return a new instance of `EdgeInsets()`.

pbartolome (approved):

.github/workflows/record-screenshots.yml
salavert: In teory yes, tried it but didn't work, maybe I'm doing something wrong

cbarcom (approved):

.github/workflows/record-screenshots.yml
alejandroruizponce: I think it would be a good idea to investigate this in a separate ticket.

Sources/Mistica/Components/Sheet/View/BottomSheetPresentation/BottomSheetInteractiveDismissalTransition.swift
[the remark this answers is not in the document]
alejandroruizponce: Missing fix, now it's fixed!

Sources/MisticaSwiftUI/Utils/Extensions/EnvironmentValues.swift
alejandroruizponce: At first I did not see how to fix it, but now I do through preconcurrency.

alexanegon (approved):

PR #414, opened 2024-11-27
IOS-10797 Change Mistica workflows to use the reusable workflows and upload builds to Firebase

pbartolome:
## 🎟️ **Jira ticket**
IOS-10797

## 🥅 **What's the goal?**
- Upload Mistica Catalog to firebase Distribution

## 🚧 **How do we do it?**
- Update GitHub Actions to use the new reusable actions. 

⚠️  As Mistica is a public repository it can't access private reusable workflows but it can checkout and run reusable composite actions. This means that this repo can't benefit of all the coordination done in the workflow and it still needs to do it itself, using the MakeFile to invoke xcodebuild ⚠️


salavert (approved):

pbartolome:
As we still want AppCenter distribution, I've added a separate action for Firebase and restored the AppCenter action

cbarcom (approved):

PR #415, opened 2024-11-28
IOS-10800 Make ViewStates component in SwiftUI

alejandroruizponce:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-10800

## 🥅 **What's the goal?**
As a performance objective to make SwiftUI components, ViewStates component will be migrated from UIKit to SwiftUI.

## 🚧 **How do we do it?**
A new component similar to the component in UIKit has been created and added inside the Catalog.

**FOR DESIGN MEMBER** (@AnaMontes11) : There is a difference between the buttons we have for SwiftUI and the UIKit buttons, I'm not clear which one should be the correct one since I can't find references to the Design of this component. I hope that in the revision of this PR a member of Design can clarify it.

If the SwiftUI design is correct, it could be used as a base and apply in UIKit a wrapper that encapsulates the SwiftUI design to use the same design in all platforms.

| SwiftUI | UIKit |
|---|---|
| Screenshot 2023-10-03 at 15 27 29 |Screenshot 2023-10-03 at 15 30 48|

## 🏑 **AppCenter build**
Screenshot 2024-11-29 at 12 12 00

### DIFF
--- a/MisticaCatalog/MisticaCatalog.xcodeproj/project.pbxproj
+++ b/MisticaCatalog/MisticaCatalog.xcodeproj/project.pbxproj
@@ -53,6 +53,7 @@
 		18E485A4287F19EB0052A6F2 /* UICatalogCroutonViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18E48575287F19EB0052A6F2 /* UICatalogCroutonViewController.swift */; };
 		18E485A5287F19EB0052A6F2 /* UICatalogBadgeViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18E48576287F19EB0052A6F2 /* UICatalogBadgeViewController.swift */; };
 		18E485A6287F19EB0052A6F2 /* UICatalogHeaderViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18E48577287F19EB0052A6F2 /* UICatalogHeaderViewController.swift */; };
+		243730502CF74F95008DECFB /* ViewStatesCatalogView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2437304F2CF74F95008DECFB /* ViewStatesCatalogView.swift */; };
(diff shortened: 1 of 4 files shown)


DavidMarinCalleja (changes requested):
There would be a missing test for this component, wouldn't there?

alejandroruizponce:
> Hi @alejandroruizponce , Here’s the link to the button component specifications from Mística: [Buttons component Specs](https://www.figma.com/design/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0-1085&t=ax5v29e42Sl4h6j7-4). You can review it to clarify any doubts and make the necessary adjustments. Let me know if this helps or if you need anything else. :)

Thanks @AnaMontes11, but really my question is more about what type of button is correct for this View State component. If the one we use in UIKit or the SwiftUI one, or if both are valid.

alejandroruizponce:
> There would be a missing test for this component, wouldn't there?

There were no tests for the component in UIKit either, if I can I will check and add some.

AnaMontes11:
Hi @alejandroruizponce, this is not a Mística component; I understand it comes from an old View State. On our side, it doesn’t matter if you use a small or large button, although for consistency, it’s better to use the same one in both. If this decision is relevant for Novum or could affect something in production, we’d need to discuss it with the product team. Let me know if I’ve clarified your question.

DavidMarinCalleja (approved):

AnaMontes11:
Hello again, @alejandroruizponce As I mentioned, it seems like an older empty state. If it helps, I would use the same (large) button in the SwiftUI version for consistency and discuss with the team updating both to something closer to a more modern empty state.

franciscojrp (approved):

alejandroruizponce:
> Hello again, @alejandroruizponce As I mentioned, it seems like an older empty state. If it helps, I would use the same (large) button in the SwiftUI version for consistency and discuss with the team updating both to something closer to a more modern empty state.

Talked with @AnaMontes11 and @aweell, same large button will be used on both platforms even though they are different right now. The unification of the button can be looked at in a different ticket as well as the update of the texts:
https://jira.tid.es/browse/IOS-10849

PR #421, opened 2025-01-13
Remove misticaCommon fix to an old Bundle.module bug

franciscojrp:
## 🥅 **What's the goal?**
In Wallet, all the previews crash due to an old fix implemented that seems to fail with the latest Xcode.
This PR removes that fix as it is the cause of the crashes and I don't see any other errors reverting it.

## 🚧 **How do we do it?**
Remove the fix and go back to using Bundle.module

The PR is done on the Swift 6 branch as Wallet is already using Swift 6, I have not tested it in main

### DIFF
--- a/Sources/MisticaCommon/Extensions/Bundle+Mistica.swift
+++ b/Sources/MisticaCommon/Extensions/Bundle+Mistica.swift
@@ -11,38 +11,10 @@ import Foundation
 // A helper to access the framework bundle
 extension Bundle {
     #if SWIFT_PACKAGE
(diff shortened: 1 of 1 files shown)


DavidMarinCalleja (approved):

amegias (approved):

salavert (approved):

dhidalgofadrique (approved):
👍

alexanegon (approved):

cbarcom (approved):

pbartolome (approved):

PR #430, opened 2025-02-24
LIFE-11584 migrate to orga runners

Yermanaco:
## 🎟️ **Jira ticket**
LIFE-11584

## 🥅 **What's the goal?**
Same reason as this [review](https://github.com/Telefonica/wallet-ios/pull/391)


pbartolome (approved):
this repo is public, in case we want to use different runners...

Yermanaco:
> this repo is public, in case we want to use different runners...

right, internal orga runners cannot be used in public repos

PR #437, opened 2025-03-14
IOS-11139 Remove crouton queue to avoid enqueuing messages. From now …

idenjoe:
## 🎟️ **Jira ticket**
[IOS-11139](https://jira.tid.es/browse/IOS-11139) Remove croutons queue behavior in UIKit
## 🥅 **What's the goal?**
Remove croutons queue from UIKit to avoid that a consecutive crouton is not shown until the previous one is closed.

## 🚧 **How do we do it?**
Remove the enque solution in CroutonController and dismiss the previous crouton if exists when a new one needs to be shown.

Open the Catalog and try to launch several UIKit croutons.

https://github.com/user-attachments/assets/a2955676-1c1d-4e32-8ae1-2f6b584cd7f6

## 🏑 **AppCenter build**

qr

### DIFF
--- a/Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
+++ b/Sources/Mistica/Components/Crouton/Presentation/CroutonController.swift
@@ -13,7 +13,7 @@ public class CroutonController: NSObject {
         public typealias Closure = () -> UIViewController?
         public static let `default`: Closure = { UIApplication.shared.windows.filter(\.isKeyWindow).first?.rootViewController }
     }
-    
+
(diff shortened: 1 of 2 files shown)


idenjoe:
Merge as a breaking change because we have removed the token we previously return when create a crouton to be able to dismiss it if it was enqueued.

amegias (approved):
cool!

dhidalgofadrique (approved):
Good job!

idenjoe:
@yceballost Remember include this change in the spec please.

PR #441, opened 2025-03-17
IOS-11147 Update releasrc and update_version script to be executed in…

idenjoe:
## 🎟️ **Jira ticket**
[IOS-11147](https://jira.tid.es/browse/IOS-11147) Force to execute update_version script in a bash shell

## 🥅 **What's the goal?**
After updating semantic release to v24.2.3 and Node to v20.8.1 in the release GA workflow it's failing when executing the update_version script with a syntax error: https://github.com/Telefonica/mistica-ios/actions/runs/13906207275/job/38909727041

I have been looking for a solution and it seems that the default shell used by a GA is sh and the script is a bash shell. [[ ... ]] seems to not be fully supported by sh and is a Bash-specific syntax. The question is why was working with the previous semantic-release/exec versions...

## 🚧 **How do we do it?**
Update releasers config file and the update_version script to force to be executed in a bash shell.

Launching the GA( the dry-run mode does skip the exec prepare stuff...)
I've create a branch to run only the update version script with both shells, sh and bash. The results( note that the error is different in both cases):
| sh | bash |
|--------|-------|
|   Screenshot 2025-03-18 at 10 14 32    |    Screenshot 2025-03-18 at 10 21 13   |

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -4,7 +4,7 @@
         "@semantic-release/release-notes-generator",
         "@semantic-release/changelog",
         ["@semantic-release/exec", {
-            "prepareCmd": "./scripts/update-version.sh ${nextRelease.version}"
+            "prepareCmd": "bash ./scripts/update-version.sh ${nextRelease.version}"
         }],
         ["@semantic-release/git", {
             "assets": ["CHANGELOG.md", "Sources/SupportFiles/Mistica.xcconfig", "MisticaCatalog/SupportFiles/MisticaCatalogConfig.xcconfig"]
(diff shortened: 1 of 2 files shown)


pbartolome:
The issue is probably that now the runner is linux (ubuntu-latest) instead of a Mac, we can change to a public Mac or update the script so it can run in any machine

amegias:
> The issue is probably that now the runner is linux (ubuntu-latest) instead of a Mac, we can change to a public Mac or update the script so it can run in any machine

My vote: I would keep running in ubuntu

amegias (approved):

idenjoe:
I don't see the need to run this script on a Mac. I would keep executing this in Ubuntu.

pbartolome (approved):

amegias:
> The issue is probably that now the runner is linux (ubuntu-latest) instead of a Mac, we can change to a public Mac or update the script so it can run in any machine

I would also keep ubuntu if possible

PR #445, opened 2025-04-07
Created TextLink component and tests

Gomay88:
## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-11039

## 🥅 **What's the goal?**
As specified in this [figma](https://www.figma.com/design/81rPCqGchGf54aoh2kpF1o/🔸-TextLink-Specs?node-id=1-75&t=lOZzLRPnIOSVBx7A-0) we want to create the textLink component

## 🚧 **How do we do it?**
Created a component which given a complete string and none, 1 word or multiple words in that string will underline them and  change the color to textLink. Also accept the inverse color with isInverse parameter

https://github.com/user-attachments/assets/aff000b0-3e4e-47b3-995e-9aebc457d550


Sources/Mistica/Components/TextLink/TextLink.swift
```diff
@@ -0,0 +1,70 @@
...
+
+public protocol TextLinkDelegate: AnyObject {
+    func textLink(_ view: TextLink, tappedText: String)
+}
+
+public class TextLink: UITextView, UITextViewDelegate {
+    public weak var linkDelegate: TextLinkDelegate?
+    private var linkedWords: [String] = []
+
+    public init(fullText: String, linkedWords: [String]? = nil, font: UIFont = .textPreset2(weight: .link), color: UIColor = .black, isInverse: Bool = false) {
```
dhidalgofadrique, line 19: I think it has no sense using a TextLink component with no links (nil or empty array). It could lead to a misuse of the component. IMO The array should be non optional and maybe add a precondition like:

`precondition(!linkedWords.isEmpty, "At least one link has to be provided")`

WDYT?

yceballost:
How could I install this app with the new distribution system?

Sources/Mistica/Components/TextLink/README.md
```diff
@@ -0,0 +1,3 @@
+#  TextLink
+
+The TextLink component find words inside of a string, marks it as underlined and use a delegate to manage where the user taps. If no word if provided the full string will act as a link and be underlined.
```
dhidalgofadrique, line 3:
```suggestion
The TextLink component finds words inside of a string, marks it as underlined and uses a delegate to manage where the user taps. If no word is provided, the full string will act as a link and will be underlined.
```

dhidalgofadrique (approved):
Great job!
respect

yceballost (approved):
image

Sources/Mistica/Components/TextLink/README.md
DavidMarinCalleja: @Gomay88   I have some doubts:
- As far as I understand this component does not open a weblink. It only marks certain parts of a text and using an indicated delegate that the user has taped.
- If I have this text "**Link** and **Link**", how can I know in the delegate on which "Link" I have taped?
-  In the same example, if I want to mark the second Link but not the first one, how can I do it? "Link and **Link**"

dhidalgofadrique: Good point, if a word appears twice, both are underlined

Captura de pantalla 2025-04-08 a las 12 48 53

MisticaCatalog/Source/Catalog/Mistica/Components/UICatalogTextLinkViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+
+private extension UICatalogTextLinkViewController {
+    var text: String? {
+        guard let text = textCell.textField.text, !text.isEmpty else { return nil }
+        return text
+    }
+
+    var links: [String]? {
+        guard let links = linkCell.textField.text, !links.isEmpty else { return nil }
+        let separatedLinks = links.components(separatedBy: ",")
```
dhidalgofadrique, line 139: Minor (only applies to catalog: Trim to avoid underline spaces for example in "Tap, here"
```suggestion
        let separatedLinks = links.components(separatedBy: ",").map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
```

dhidalgofadrique (approved):
goodJobEddieMurphy

Sources/Mistica/Components/TextLink/README.md
Gomay88: Done and updated the video

Tests/MisticaTests/UI/TextLinkTests.swift
```diff
@@ -0,0 +1,47 @@
...
+//  Made with ❤️ by Novum
+//
+//  Copyright © Telefonica. All rights reserved.
+//
+
+@testable import Mistica
+import SnapshotTesting
+import XCTest
+
+final class TextLinkTests: XCTestCase {
```
L-Trujillo26, line 13: it might be useful to add tests for invalid inputs, such as empty text or nil values (using just one brand to verify there aren't errors)

DavidMarinCalleja: +1

Tests/MisticaTests/UI/TextLinkTests.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    func testSingleWordTextLinkView() {
+        assertSnapshotForAllBrandsAndStyles(
+            as: .image(size: CGSize(width: 200.0, height: 33.0)),
+            viewBuilder: TextLink(fullText: "Text 1 link word", linkedWords: [LinkedWord(word: "link")])
+        )
+    }
+
+    func testMultipleWordTextLinkView() {
+        assertSnapshotForAllBrandsAndStyles(
```
DavidMarinCalleja, line 28: maybe it was not necessary to do it for all Brands, since we only want to test behaviours with different values.

Sources/Mistica/Components/TextLink/TextLink.swift
```diff
@@ -0,0 +1,107 @@
...
+
+public class TextLink: UITextView, UITextViewDelegate {
+    public weak var linkDelegate: TextLinkDelegate?
+
+    private var linkedWords: [LinkedWord] = []
+    private var linkRanges: [NSRange: String] = [:]
+
+    // Flag to prevent voiceOver bug to call delegate twice
+    private var isHandlingTap = false
+    private let debounceInterval: TimeInterval = 1
```
Gomay88, line 33: As discussed with @dhidalgofadrique there is a bug when there are 2 or more links in the same UITextView where the delegate `shouldInteractWith` get called twice so I added a flag and added 1 second to change that flag to prevent this delegate to be called twice(In the new video you'll see it only get call once)

Sources/Mistica/Components/TextLink/TextLink.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        attributedText = attributed
+    }
+
+    public func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        guard URL.scheme == "textLink" else { return true }
+
+        guard !isHandlingTap else { return false }
```
DavidMarinCalleja, line 93: note that access to `isHandlingTap` is not synchronised, i.e. if two threads enter this line at the same time they will both read false and therefore the `guard` will be enforced and two threads will both try to execute code at the same time.

Tests/MisticaTests/UI/TextLinkTests.swift
Gomay88: Still need to test for the colors in different brands right?

Sources/Mistica/Components/TextLink/TextLink.swift
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        guard URL.scheme == "textLink" else { return true }
+        
+        if UIAccessibility.isVoiceOverRunning && isHandlingTap {
+            isHandlingTap = false
+            return false
+        }
+        isHandlingTap = true
```
Gomay88, line 95: Instead of last commit using a delay which can be a problem in the future, thx to @DavidMarinCalleja we discard the second time the delegate is called when we are using VoiceOver

dhidalgofadrique: But with this solution if the user with VoiceOver enabled taps a link that, for example, opens a new window, then closes the window and taps the same link... it would ignore that second tap, isn't it?

Gomay88: Shouldnt be like that. With voiceOver when the user taps a link will call 2 times to this delegate so the first time will open the link and the second one will change isHandlingTap to false so when the user taps it another time it will work as expected

dhidalgofadrique: Ok, however we have a risk: If Apple (by any reason 😅) fixes this bug and it calls only once, it will ignore the second call...

But there is no a "good" solution in my opinion...

PR #446, opened 2025-04-22
IOS-10545 Fix Strict Concurrency in XCode 16.2

alejandroruizponce:
## 🎟️ **Jira ticket**

## 🥅 **What's the goal?**
Concurrency bugs have been detected that only appear in XCode version 16.2. 

The fact that they do not appear in the other versions of XCode does not mean that it is a concurrency error that must be fixed.

## 🚧 **How do we do it?**
The closure of the `onPreferenceChange` method is `Sendable` and therefore if we want to access an isolated value in `MainActor` we must guarantee that the context where it is accessed is also `MainActor`.

However, `onPreferenceChange` method is not isolated in `MainActor` so it is necessary to isolate its closure in a `Task` that is.

You should not see any compile errors in XCode 16.2.


dhidalgofadrique (approved):
Thanks for fix it!

DavidMarinCalleja (approved):

PR #451, opened 2025-05-28
Add flow token to notify

pbartolome:
In order to notify to teams in public runners using the internal action we need the private novum token

This is to fix release executions like: 
https://github.com/Telefonica/mistica-ios/actions/runs/15303623918/job/43050533903#step:6:34


dhidalgofadrique (approved):
Thanks for fixing it!

PR #452, opened 2025-06-06
Ios 11095 automatic focus snackbar critical cases

L-Trujillo26:
> [!WARNING]
> Should a designer need to review/verify this PR?

## 🎟️ **Jira ticket**
https://jira.tid.es/browse/IOS-11095

## 🥅 **What's the goal?**
The Snackbar specifications have been updated for accessibility. For critical persistent Snackbars, the focus will now automatically move to ensure better accessibility.

## 🚧 **How do we do it?**

A new parameter has been added to move the focus to persistent Snackbars. Additionally, a delay parameter has been introduced to ensure the focus movement, for cases when UI changes are still rendering or have animations.

https://github.com/user-attachments/assets/25bf7ad5-edc8-43d3-8c8d-ec97ca598022

### DIFF
--- a/Sources/Mistica/Components/Snackbar/Presentation/SnackbarView.swift
+++ b/Sources/Mistica/Components/Snackbar/Presentation/SnackbarView.swift
@@ -211,7 +211,7 @@ extension SnackbarView {
                 container.layoutIfNeeded()
             },
             completion: { _ in
-                AccessibilityHelper.post(self.text)
+                self.triggerAccessibilityAction()
 
                 container.clipsToBounds = previousClipsToBounds
 
@@ -267,6 +267,13 @@ extension SnackbarView {
 private extension SnackbarView {
     var shouldShowCloseButton: Bool {
         forceDismiss || (action == nil && config.overrideDismissInterval == .infinite(nil))
+    }
+    func triggerAccessibilityAction() {
+        if config.overrideDismissInterval.shouldAutoFocus {
+            UIAccessibility.post(notification: .layoutChanged, argument: self)
+        } else {
+            AccessibilityHelper.post(self.text)
+        }
     }
 
     func layoutViews() {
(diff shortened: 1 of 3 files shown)


Sources/Mistica/Components/Snackbar/Presentation/SnackbarView.swift
```diff
@@ -269,6 +270,18 @@ private extension SnackbarView {
...
+            let delay = config.overrideDismissInterval.autoFocusDelay ?? 0.0
+            DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
+                UIAccessibility.post(notification: .layoutChanged, argument: self)
+            }
+
+        } else {
+            AccessibilityHelper.post(text)
+        }
+    }
+
```
dhidalgofadrique, line 284: AFAIK if the focus is moved to a view, the VoiceOver will speak the text of the view being focused. In our case if this view is the snackbar and the first view is the label it should speak it automatically, isn't it?

dhidalgofadrique:
Remeber that Mistica has UIKit and SwiftUI, you should replicate this behavior also here: https://github.com/Telefonica/mistica-ios/blob/7a0e739a7acdf5c01df4ac0cd221d31508f756ba/Sources/MisticaSwiftUI/Components/Snackbar/Snackbar.swift#L101

L-Trujillo26:
> Remeber that Mistica has UIKit and SwiftUI, you should replicate this behavior also here:
> 
> https://github.com/Telefonica/mistica-ios/blob/7a0e739a7acdf5c01df4ac0cd221d31508f756ba/Sources/MisticaSwiftUI/Components/Snackbar/Snackbar.swift#L101

UIAccessibility requires a UIView to move focus, which is not compatible with SwiftUI.
From iOS 15, SwiftUI provides the native @AccessibilityFocusState modifier for focus management.
Currently this functionality is not required for any SwiftUI components in this project.
If it becomes necessary while still supporting iOS versions below 15, a temporary compatible solution (UIKit  wrapper and modifiers) should be implemented.

dhidalgofadrique (approved):
goodJob10

